PR #35, opened 2015-10-26
Added threshold, otsu level

akonneker:
I spent some wasted time chasing a nonexistent bug trying to get the results to match matlab's... but it should work now. Haven't used it on actual images yet, but I will "soon".

The style may be terrible, but I am a rank amateur when it comes to rust, so I hope you'll forgive me.

I'll be swamped for a little while yet, but when I can carve out time I'm going to port some matlab code for more interesting algorithms.

I'll also add some other thresholding options eventually, too.

### DIFF
--- a/src/contrast.rs
+++ b/src/contrast.rs
@@ -10,7 +10,9 @@ use image::{
 
 /// Returns the Otsu threshold level of an 8bpp image.
 /// This threshold will optimally binarize an image that
-/// meets certain criteria.
+/// contains two classes of pixels which have distributions
+/// with equal variances. For details see:
+/// Xu, X., et al. Pattern recognition letters 32.7 (2011)
 pub fn otsu_level<I>(image: &I) -> u8
     where I: GenericImage<Pixel=Luma<u8>> {
     let hist = histogram(image);
@@ -29,8 +31,8 @@ pub fn otsu_level<I>(image: &I) -> u8
     }
 
     let mut sigma_max:f64 = -1.0;
-    let mut otsu:f64 = 0.0;
-    let mut otsu2:f64 = 0.0;
+    let mut otsu = 0f64;
+    let mut otsu2 = 0f64;
 
     for i in 0..levels {
         meanc[i] /= pixel_count;
@@ -57,16 +59,11 @@ pub fn otsu_level<I>(image: &I) -> u8
 /// obtained by applying the given image
 pub fn threshold<I>(image: &I, thresh: u8) -> GrayImage
     where I: GenericImage<Pixel=Luma<u8>> {
+
     let mut out: GrayImage = ImageBuffer::new(image.width(), image.height());
-     for y in 0..image.height() {
-            for x in 0..image.width() {
-                if image.get_pixel(x, y)[0] as u8 <= thresh {
-                    out.put_pixel(x, y, Luma([0 as u8]));
-                } else {
-                    out.put_pixel(x, y, Luma([255 as u8]));
-                }
-            }
-        }
+    out.copy_from(image, 0, 0);
+    threshold_mut(&mut out, thresh);
+
     out
 }
 
@@ -77,9 +74,9 @@ pub fn threshold_mut<I>(image: &mut I, thresh: u8)
      for y in 0..image.height() {
             for x in 0..image.width() {
                 if image.get_pixel(x, y)[0] as u8 <= thresh {
-                    image.put_pixel(x, y, Luma([0 as u8]));
+                    image.put_pixel(x, y, Luma([0]));
                 } else {
-                    image.put_pixel(x, y, Luma([255 as u8]));
+                    image.put_pixel(x, y, Luma([255]));
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/contrast.rs
```diff
@@ -7,9 +8,86 @@ use image::{
     Luma
 };
 
-/// Returns the cumulative histogram of grayscale values in an 8bpp
+/// Returns the Otsu threshold level of an 8bpp image.
+/// This threshold will optimally binarize an image that
+/// meets certain criteria.
+pub fn otsu_level<I>(image: &I) -> u8
```
theotherphil: "meets certain criteria". Which criteria?

src/contrast.rs
```diff
@@ -7,9 +8,86 @@ use image::{
...
+    for i in 1..levels {
+        pixel_count += hist[i] as f64;
+        histc[i] = histc[i-1] + hist[i];
+        meanc[i] = meanc[i-1] + (hist[i] as f64)*(i as f64);
+    }
+
+    let mut sigma_max:f64 = -1.0;
+    let mut otsu:f64 = 0.0;
+    let mut otsu2:f64 = 0.0;
+
```
theotherphil: Minor nitpick: most rust uses a space after the colons here. Also, you could just write let mut x = 0f64.

src/contrast.rs
```diff
@@ -7,9 +8,86 @@ use image::{
...
+    }
+    ( (otsu + otsu2) / 2.0 ).ceil() as u8
+}
+
+/// Returns a binarized image from an input 8bpp grayscale image
+/// obtained by applying the given image
+pub fn threshold<I>(image: &I, thresh: u8) -> GrayImage
+    where I: GenericImage<Pixel=Luma<u8>> {
+    let mut out: GrayImage = ImageBuffer::new(image.width(), image.height());
+     for y in 0..image.height() {
```
theotherphil: I suspect you can get away without the type annotation here.

src/contrast.rs
```diff
@@ -7,9 +8,86 @@ use image::{
...
+            for x in 0..image.width() {
+                if image.get_pixel(x, y)[0] as u8 <= thresh {
+                    out.put_pixel(x, y, Luma([0 as u8]));
+                } else {
```
theotherphil: And here.

src/contrast.rs
```diff
@@ -7,9 +8,86 @@ use image::{
...
+                    out.put_pixel(x, y, Luma([255 as u8]));
+                }
```
theotherphil: None of these "as u8" are necessary.

src/contrast.rs
```diff
@@ -7,9 +8,86 @@ use image::{
...
+            }
+        }
+    out
+}
+
+/// Mutates given image to form a binarized version produced by applying
+/// the given threshold
+pub fn threshold_mut<I>(image: &mut I, thresh: u8)
+    where I: GenericImage<Pixel=Luma<u8>>  {
+     for y in 0..image.height() {
```
theotherphil: When there's a pair of f and f_mut, I'd prefer that f_mut contain the full logic, and the body of f just be: "let out = ImageBuffer::new(...); out.copy_from(input, 0, 0); f(out); out". See e.g equalize_histogram and equalize_histogram_mut.

src/contrast.rs
```diff
@@ -83,6 +175,51 @@ mod test {
...
+                255u8,  255u8,  255u8,  255u8]).unwrap();
+
+        let test = threshold(&before, 125u8);
+
+        for y in 0..after.height() {
+            for x in 0..after.width() {
+                let check = after.get_pixel(x, y)[0] as usize;
+                let test_pixel = test.get_pixel(x, y)[0] as usize;
+                assert_eq!(test_pixel, check);
+            }
```
theotherphil: assert_pixels_eq will do this more succinctly: assert_pixels_eq!(test, after). I'd use "actual" for "test" and "expected" for "after" here, as it's slightly clearer what each value is for.

theotherphil:
Looks good, thanks! I've added a selection of pretty pedantic comments - I hope you don't mind the nitpicking. Nothing major, so I'll merge this either when you make the changes or in a couple of days if you don't have the time/inclination to make them.

theotherphil:
Any other stuff you fancy adding would be great. If the other algorithms you're considering porting are "interesting" in non-image-processing ways then maybe we should put them in a currently nonexistent computer vision sister project.

akonneker:
I'll happily make the changes suggested. As I said, my rust experience is essentially nil, so I appreciate any suggestions. 

What I'm going to implement first are some things I'm already working with/on. One is a [phase congruency filter] (http://cvpr11.cecs.anu.edu.au/files/PhaseCongCVPR.pdf). This would be a nice addition to the library, I think.

The other is an RBG-D SLAM algorithm like [kinectfusion](https://github.com/Nerei/kinfu_remake) or [dvo-slam](https://github.com/tum-vision/dvo_slam).

So SLAM would definitely need a new home. I'll start on that in earnest pretty soon, but it'll be a few weeks at the earliest before it's finished. So I'll let you know.

theotherphil:
I'd not heard of phase congruency filters before - looks interesting. I'm also planning on implementing a few vision algorithms soon (http://www.edwardrosten.com/work/taylor_2009_robust.pdf, https://www.cs.cmu.edu/~efros/courses/LBMV07/Papers/viola-cvpr-01.pdf) but it doesn't look like there'll be much overlap in required code yet.

src/contrast.rs
akonneker: These annotations appear to be necessary to avoid type mismatch errors.

theotherphil:
Great, thanks!

PR #65, opened 2016-01-02
simpler has_dark/bright_span

tafia:
Avoid nested loops, iterate only once on circle values.

On the bench in my system: 
```
before
test corners::test::bench_is_corner_fast9_9_contiguous_lighter_pixels ... bench:          59 ns/iter (+/- 7)

after
test corners::test::bench_is_corner_fast9_9_contiguous_lighter_pixels ... bench:          40 ns/iter (+/- 5)
```

not really significant but still an improvement

### DIFF
--- a/src/corners.rs
+++ b/src/corners.rs
@@ -5,8 +5,6 @@ use image::{
     Luma
 };
 
-use std::cmp;
-
 /// A location and score for a detected corner.
 /// The scores need not be comparable between different
 /// corner detectors.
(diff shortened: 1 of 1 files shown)


theotherphil:
Thanks, that's a substantial speed up! The new code panics if length > 16, but the old code didn't handle this case properly either. As these functions are private it seems reasonable to just trust the caller.

tafia:
Thanks!
Indeed, forgot to mention about `unwrap`...

PR #67, opened 2016-01-03
Corners

tafia:
Some more optimization on corner bright/dark corner check.
1- use a generic function `search_span`
2- use `u8` instead of `usize` for length
3- early return `false` if `length > 16` (no impact on perf and avoids a panic)
4- count items not matching condition: as length is bigger than 16/2, nb_ko will probably reach 0 before nb_ok reaches length.

```
test corners::test::bench_is_corner_fast12_12_noncontiguous           ... bench:          31 ns/iter (+/- 3)
test corners::test::bench_is_corner_fast9_9_contiguous_lighter_pixels ... bench:          30 ns/iter (+/- 1)
```


tafia:
@theotherphil I can merge it myself but I'd prefer having one review.

tafia:
closed as 4. doesn't seam to be true.
I've done another PR #68 that I merged directly as it is simpler

PR #68, opened 2016-01-03
Corners

tafia:
more dark/bright corner changes

- early return if length > 16
- length as u8
- use a generic function search_span
- add a bench for noncontiguous


theotherphil:
Thanks!

PR #80, opened 2016-01-09
more generic filter_kernel

tafia:
Add a `filter_kernel` fn which generalizes `filter3x3`, `horizontal_filter` and `vertical_filter`.
While result are way better for horizontal and vertical there is a regression on 3x3 filter.
For the moment I've decided to generalize 3x3 but I could revert that particular function if needed.

```
before
test filter::test::bench_filter3x3_i32_filter                         ... bench:   3,761,471 ns/iter (+/- 92,914)
test filter::test::bench_horizontal_filter                            ... bench:  10,581,024 ns/iter (+/- 224,687)
test filter::test::bench_separable_filter                             ... bench:   7,518,389 ns/iter (+/- 174,294)
test filter::test::bench_vertical_filter                              ... bench:  10,785,773 ns/iter (+/- 121,640)

after
test filter::test::bench_filter3x3_i32_filter                         ... bench:   4,183,313 ns/iter (+/- 351,771)
test filter::test::bench_horizontal_filter                            ... bench:   4,891,864 ns/iter (+/- 383,548)
test filter::test::bench_separable_filter                             ... bench:   3,575,458 ns/iter (+/- 110,890)
test filter::test::bench_vertical_filter                              ... bench:   5,033,250 ns/iter (+/- 189,755)
```

### DIFF
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -86,6 +86,70 @@ pub fn box_filter<I>(image: &I, x_radius: u32, y_radius: u32)
     out
 }
 
+pub struct Kernel<'a, K: 'a> {
+    data: &'a [K],
+    width: u32,
+    height: u32,
+}
+
+impl<'a, K: Num + Copy + 'a> Kernel<'a, K> {
+
+    pub fn new(data: &'a [K], width: u32, height: u32) -> Kernel<'a, K> {
+        assert!(width * height == data.len() as u32,
+            format!("Invalid kernel len: expecting {}, found {}",
+                    width * height, data.len()));
+        Kernel {
+            data: data,
+            width: width,
+            height: height,
+        }
+    }
+
+    /// Returns 2d correlation of an image. Intermediate calculations are performed 
+    /// at type K, and the results converted to pixel Q via f. Pads by continuity.
+    pub fn filter<I, F, Q>(&self, image: &I, mut f: F) -> VecBuffer<Q>
+        where I: GenericImage,
+              <I::Pixel as Pixel>::Subpixel: ValueInto<K>,
+              Q: Pixel + 'static,
+              F: FnMut(&mut Q::Subpixel, K) -> (),
+    {
+
+        let (width, height) = image.dimensions();
+        let mut out = VecBuffer::<Q>::new(width, height);
+        let num_channels = I::Pixel::channel_count() as usize;
+        let zero = K::zero();
+        let mut acc = vec![zero; num_channels];
+        let (k_width, k_height) = (self.width, self.height);
+
+        for y in 0..height {
+            for x in 0..width {
+                for k_y in 0..k_height {
+                    let y_p = cmp::min(height + height - 1, 
+                                       cmp::max(height, (height + y + k_y - k_height / 2))) - height;
+                    for k_x in 0..k_width {
+                        let x_p = cmp::min(width + width - 1, 
+                                           cmp::max(width, (width + x + k_x - k_width / 2))) - width;
+                        let (p, k) = unsafe {
+                            (image.unsafe_get_pixel(x_p, y_p),
+                             *self.data.get_unchecked((k_y * k_width + k_x) as usize))
+                        };
+                        accumulate(&mut acc, &p, k);
+                    }
+                }
+                let out_channels = out.get_pixel_mut(x, y).channels_mut();
+                for (a, c) in acc.iter_mut().zip(out_channels.iter_mut()) {
+                    f(c, a.clone());
+                    *a = zero;
+                }
+            }
+        }
+
+        out
+    }
+
+
+}
+
 /// Returns 2d correlation of view with the outer product of the 1d
 /// kernels h_filter and v_filter.
 pub fn separable_filter<I, K>(image: &I, h_kernel: &[K], v_kernel: &[K])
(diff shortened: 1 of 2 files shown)


src/filter.rs
```diff
@@ -105,152 +105,90 @@ pub fn separable_filter<I, K>(image: &I, h_kernel: &[K], v_kernel: &[K])
...
+          <I::Pixel as Pixel>::Subpixel: ValueInto<K>,
+          Q: Pixel + 'static,
+          F: FnMut(&mut Q::Subpixel, K) -> (),
           K: Num + Copy {
 
-    let (width, height) = image.dimensions();
-    let mut out = ImageBuffer::<ChannelMap<P, S>, Vec<S>>::new(width, height);
-    let num_channels = I::Pixel::channel_count() as usize;
+    assert!(kernel.len() as u32 >= k_width * k_height,
+            "incompatible kernel lengths".to_owned());
```
theotherphil: It doesn't feel right to have to pass in width and height separately from the kernel. What's the use case for allowing kernel length > width * height? 

We could introduce a trivial Kernel struct just containing an array (slice?), width and height.

src/filter.rs
```diff
@@ -105,152 +105,90 @@ pub fn separable_filter<I, K>(image: &I, h_kernel: &[K], v_kernel: &[K])
...
-                 &image.unsafe_get_pixel(x     , y     ),
-                 &image.unsafe_get_pixel(x_next, y     ),
-                 &image.unsafe_get_pixel(x_prev, y_next),
-                 &image.unsafe_get_pixel(x     , y_next),
-                 &image.unsafe_get_pixel(x_next, y_next),]
-            };
-            for i in 0..9 {
-                accumulate(&mut acc, pixels[i], kernel[i]);
+    let (w, h) = image.dimensions();
+    let mut out = VecBuffer::<Q>::new(w, h);
```
theotherphil: This might sound pedantic, but I think we should produce a style guide on names for standard sorts of things. e.g. num_things vs nb_things vs no_things, w vs width, h vs height, etc., etc.

Obviously any choice is fairly arbitrary, and I don't think we need to enforce this religiously, but I always find it easier/more pleasant to read codebases that are consistent. I'm not suggesting that you need to make any changes to this pull request for this reason, but I'll probably do a sweep through the codebase renaming things when we agree on what the conventions should be.

src/filter.rs
```diff
@@ -105,152 +105,90 @@ pub fn separable_filter<I, K>(image: &I, h_kernel: &[K], v_kernel: &[K])
...
+/// performed at type K, and the results clamped to subpixel type S. Pads by continuity.
+pub fn filter3x3<I, P, K, S>(image: &I, kernel: &[K]) -> VecBuffer<ChannelMap<P, S>>
+    where I: GenericImage<Pixel=P>,
+          P::Subpixel: ValueInto<K>,
+          S: Clamp<K> + Primitive + 'static,
+          P: WithChannel<S> + 'static,
+          K: Num + Copy {
+    filter_kernel(image, kernel, 3, 3, |channel, acc| *channel = S::clamp(acc))
+}
+
```
theotherphil: I'm not too fussed about the performance regression for this function - I suspect there are still performance improvements to be had in your new filter_kernel function, so writing a specialisation for filter3x3 can wait until we're reasonably sure that filter_kernel is about as fast as it's going to get.

theotherphil:
Largely LGTM, except for passing kernel dimensions separately to the slice - seems like an easily avoidable source of potential bugs. Thanks!

tafia:
No pb
I'll create a dedicated struct for handling kernel with its dimensions.

src/filter.rs
tafia: Speaking of this, do you know if we could find such benchmarks somewhere ? 
I know benchmark is a complicated topic but I have no clue how slow we are now compared to other libraries.

tafia:
Rebased and added the kernel struct

tafia:
Note: Results are actually even better when forcing inline on image `unsafe_get_pixel` and `unsafe_put_pixel`.

Corresponding PR: https://github.com/PistonDevelopers/image/pull/505

```
before with inline
test filter::test::bench_filter3x3_i32_filter                         ... bench:   2,466,759 ns/iter (+/- 187,251)
test filter::test::bench_horizontal_filter                            ... bench:  10,729,812 ns/iter (+/- 1,679,107)
test filter::test::bench_separable_filter                             ... bench:   7,823,963 ns/iter (+/- 746,706)
test filter::test::bench_vertical_filter                              ... bench:  10,440,924 ns/iter (+/- 1,344,521)

after with inline
test filter::test::bench_filter3x3_i32_filter                         ... bench:   2,731,488 ns/iter (+/- 25,606)
test filter::test::bench_horizontal_filter                            ... bench:   2,968,470 ns/iter (+/- 137,654)
test filter::test::bench_separable_filter                             ... bench:   2,670,435 ns/iter (+/- 45,366)
test filter::test::bench_vertical_filter                              ... bench:   4,449,541 ns/iter (+/- 49,640)
```

theotherphil:
Thanks!

src/filter.rs
theotherphil: Sadly, I'm not aware of any existing benchmarks. My plans didn't extend beyond just installing/building some other libraries and writing the same benchmarks we have in this one. I've not yet done this.

PR #102, opened 2016-01-13
Faster connected_components

theotherphil:
At the cost of using potentially twice as much memory for images with few connected components.

Huge speed up on the benchmarks, but these are fairly small images. I think it's a net win. Even if we decide this isn't the best approach, it's definitely better than using hash maps like the code currently does.

```
Before:

test regionlabelling::test::bench_connected_components_eight_chessboard ... bench:   2,194,672 ns/iter (+/- 249,447)
test regionlabelling::test::bench_connected_components_four_chessboard  ... bench:   6,433,376 ns/iter (+/- 1,290,848)

After: 

test regionlabelling::test::bench_connected_components_eight_chessboard ... bench:   1,270,101 ns/iter (+/- 119,050)
test regionlabelling::test::bench_connected_components_four_chessboard  ... bench:     772,213 ns/iter (+/- 110,943)
```

### DIFF
--- a/src/regionlabelling.rs
+++ b/src/regionlabelling.rs
@@ -10,10 +10,13 @@ use definitions::{
     VecBuffer
 };
 
-use unionfind::DisjointSetForest;
+use unionfind::{
+    DisjointSetForest
+};
 
-use std::cmp;
-use std::collections::HashMap;
+use std::{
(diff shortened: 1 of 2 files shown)


tafia:
Nice!

PR #103, opened 2016-01-14
update dependencies

tafia:
fix issues appeared in [discussion](https://github.com/PistonDevelopers/imageproc/issues/97#issuecomment-170557464)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,13 +8,13 @@ readme = "README.md"
 homepage = "https://github.com/PistonDevelopers/imageproc"
 
 [dependencies]
-conv = "0.3.1"
-image = "0.3.14"
-itertools = "0.4.2"
-nalgebra = "0.3.0"
-num = "0.1.27"
-quickcheck = "0.2.24"
-rand = "0.3.11"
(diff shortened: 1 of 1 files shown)


tafia:
Kept 2 levels only for versions. I hope crate owners will upgrade at least 2nd level if there is a breaking change.

bvssvni:
The "z" part in "x.y.z" is interpreted by Cargo as the minimum version. Change in the first non-zero number counts as a breaking change. When bumping the first non-zero number in any dependency it is a good idea to bump the first non-zero number in the library, to avoid breaking existing code.

bvssvni:
You can set "image" to 0.6.1 to get the bug fixes and bump "imageproc" to 0.2.0.

theotherphil:
Thanks both for looking into this. I won't have time to do anything tonight, but I'm happy for whatever version changes @bvssvni suggests. I'll fix on the weekend if it's not already done by then.

bvssvni:
Opened https://github.com/PistonDevelopers/imageproc/pull/104. Closing this.

PR #112, opened 2016-01-17
Contrast

tafia:
Use `VecBuffer` instead of `GenericImage`.
Added some more benchmarks (we may cherry-pick them on master ?).

```
before
test contrast::test::bench_threshold                                    ... bench:   1,175,968 ns/iter (+/- 12,929)
test contrast::test::bench_threshold_mut                                ... bench:     341,623 ns/iter (+/- 4,349)
test contrast::test::bench_equalize_histogram                           ... bench:   2,227,185 ns/iter (+/- 20,862)
test contrast::test::bench_equalize_histogram_mut                       ... bench:   1,432,186 ns/iter (+/- 33,085)

after
test contrast::test::bench_threshold                                    ... bench:      31,795 ns/iter (+/- 437)
test contrast::test::bench_threshold_mut                                ... bench:      12,227 ns/iter (+/- 274)
test contrast::test::bench_equalize_histogram                           ... bench:     948,128 ns/iter (+/- 11,067)
test contrast::test::bench_equalize_histogram_mut                       ... bench:     929,868 ns/iter (+/- 12,838)
```

### DIFF
--- a/src/contrast.rs
+++ b/src/contrast.rs
@@ -1,6 +1,6 @@
 //! Functions for manipulating the contrast of images.
 
-use image::{GenericImage, GrayImage, ImageBuffer, Luma};
+use image::{GrayImage, Luma};
 use definitions::VecBuffer;
 
 /// Returns the Otsu threshold level of an 8bpp image.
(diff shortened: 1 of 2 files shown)


tafia:
... and no more unsafe in contrast!

theotherphil:
... and much shorter code!

Only iter(), iter_mut() and clone() required for our hypothetical new image trait so far. The one slight worry I have is that the GenericImage already has a pixels_mut method, but it's marked as deprecated with some comments about lifetime issues and this not being possible in safe rust. We'll have to check whether the same issues cause problems for us.

theotherphil:
Thanks! Cherry-picking the new benchmarks sounds good too.

theotherphil:
I've not read it yet, but this thread discusses the problem: http://stackoverflow.com/questions/25730586/returning-mutable-references-from-an-iterator/25748645#25748645

PR #119, opened 2016-02-23
Implement canny edge detection algorithm.

nicokoch:
This commit adds a new module `edges`, which currently only contains the
canny edge detection algorithm. The provided implementation is heavily
based on the implementation proposed by the corresponding wikipedia
article (https://en.wikipedia.org/wiki/Canny_edge_detector), with some
simple optimizations.


src/edges.rs
```diff
@@ -0,0 +1,131 @@
...
+                low_threshold: u8,
+                high_threshold: u8)
+                -> ImageBuffer<Luma<u8>, Vec<<Luma<u8> as Pixel>::Subpixel>>
+    where I: GenericImage<Pixel = Luma<u8>> + 'static
+{
+    if high_threshold < low_threshold {
+        panic!("Low threshold is greater than high threshold: {}; {}",
+               low_threshold,
+               high_threshold)
+    }
```
theotherphil: Slightly better to assert here, I think.

src/edges.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// Runs the canny edge detection algorithm on the provided `ImageBuffer`.
+///
+/// Returns a binary image, where edge pixels have a value of 255 and non-edge pixels a value of 0.
+pub fn canny<I>(image: &I,
+                low_threshold: u8,
+                high_threshold: u8)
+                -> ImageBuffer<Luma<u8>, Vec<<Luma<u8> as Pixel>::Subpixel>>
+    where I: GenericImage<Pixel = Luma<u8>> + 'static
+{
```
theotherphil: Could you please add an explanation of what high and low threshold mean here, so that it appears in the generated docs.

nicokoch:
This implementation can still be optimized heavily by using the unsafe pixel access methods and possibly doing more operations in-place (not sure how far we can push this while still remaining correct though)

src/edges.rs
```diff
@@ -0,0 +1,131 @@
...
+use gradients::{vertical_sobel, horizontal_sobel};
+
+/// Runs the canny edge detection algorithm on the provided `ImageBuffer`.
+///
+/// Returns a binary image, where edge pixels have a value of 255 and non-edge pixels a value of 0.
+pub fn canny<I>(image: &I,
+                low_threshold: u8,
+                high_threshold: u8)
+                -> ImageBuffer<Luma<u8>, Vec<<Luma<u8> as Pixel>::Subpixel>>
+    where I: GenericImage<Pixel = Luma<u8>> + 'static
```
theotherphil: The return type can be written more concisely as ImageBuffer<Luma<u8>, Vec<u8>>.

src/edges.rs
```diff
@@ -0,0 +1,131 @@
...
+    let blurred = blur(image, SIGMA);
+
+    // 2. Intensity of gradients.
+    let gx = horizontal_sobel(&blurred);
+    let gy = vertical_sobel(&blurred);
+    let mut g = ImageBuffer::from_fn(image.width(), image.height(), |x, y| {
+        let g = (gx[(x, y)][0] as f32).hypot(gy[(x, y)][0] as f32);
+        Luma { data: [g as u8] }
+    });
+
```
theotherphil: gradients::sobel_gradients does the same as this, except that its return type is Luma<u16> to avoid the possibility of overflow.

src/edges.rs
```diff
@@ -0,0 +1,131 @@
...
+    // 4. Hysteresis to filter out edges based on thresholds.
+    hysteresis(&g, low_threshold, high_threshold)
+}
+
+/// Finds local maxima to make the edges thinner.
+fn non_maximum_suppression(g: &mut ImageBuffer<Luma<u8>, Vec<<Luma<u8> as Pixel>::Subpixel>>,
+                           gx: &ImageBuffer<Luma<i16>, Vec<<Luma<i16> as Pixel>::Subpixel>>,
+                           gy: &ImageBuffer<Luma<i16>, Vec<<Luma<i16> as Pixel>::Subpixel>>) {
+    const RADIANS_TO_DEGREES: f32 = 57.2958;
```
theotherphil: As above, the types here can be written as ```ImageBuffer<Luma<u8>, Vec<u8>>```. Or even ```VecBuffer<Luma<u8>>```, but I'm not sure if that's a typedef that's worth keeping around.

src/edges.rs
nicokoch: I don't think I can use this here, since later (in `non_maximum_suppression()`) I need `gx` and `gy` seperated to calculate the edge angles.

src/edges.rs
```diff
@@ -0,0 +1,131 @@
...
+            };
+            let pixel = *g.get_pixel(x, y);
+            // If the pixel is not a local maximum, suppress it.
+            if pixel[0] < cmp1[0] || pixel[0] < cmp2[0] {
+                g.put_pixel(x, y, Luma { data: [0] });
+            }
+        }
+    }
+}
+
```
theotherphil: There's some existing code for non-maximum suppression in the suppress module, but sadly it doesn't look to be directly applicable here.

src/edges.rs
```diff
@@ -0,0 +1,131 @@
...
+// Filter out edges with the thresholds.
+fn hysteresis(g: &ImageBuffer<Luma<u8>, Vec<<Luma<u8> as Pixel>::Subpixel>>,
+              low_thresh: u8,
+              high_thresh: u8)
+              -> ImageBuffer<Luma<u8>, Vec<<Luma<u8> as Pixel>::Subpixel>> {
+    const MAX_BRIGHTNESS: Luma<u8> = Luma { data: [255u8] };
+    const MIN_BRIGHTNESS: Luma<u8> = Luma { data: [0u8] };
+    // Init output image as all black.
```
theotherphil: If you import definition::{HasWhite,HasBlack} these are available as black() and white().

theotherphil:
I've added a few comments, all very minor. Looks good, thanks again.

It would be great if you could add a test and a benchmark for this (the test could just be a regression test that shows that the results look plausible. See tests/regression.rs for some examples). The filter module contains some example benchmarks.

src/edges.rs
theotherphil: Good point. The underlying gradients method would do the trick, if it were made public (```gradients(image, &HORIZONTAL_SOBEL, &VERTICAL_SOBEL)```), but that's not a big deal.

nicokoch:
Will address them shortly :)
This was my first time implementing canny and can probably be done a little faster, maybe a task for a future contributor.

theotherphil:
The first implementation is the hardest part! With a benchmark and regression test (and possibly some unit tests) others can easily tinker with the implementation to find some speedups.

theotherphil:
A simple unit test to check for handling overflow correctly would be a solid square inside a larger solid square of a different colour (e.g. black vs white).

src/edges.rs
nicokoch: Oh yeah, I see the problem now. Would it be a possible workaround to use u8::MAX if g > 255?

The values are all positive, since we use the hypot function.

theotherphil: That seems reasonable.

nicokoch: Ok, I fixed this. Casting this to u8 was actually only done so I can save it for debugging purposes.
We can just keep g around as an f32.

nicokoch:
Ok I'm having trouble with the regression test. Should I just use `assert_pixels_equal!()` for a pregenerated truth image?

theotherphil:
Yes, the idea with those tests is just to confirm that the function does something visually sensible and isn't silently broken in a future commit. There aren't too many of these tests because it's usually possible to write some standard unit tests instead. So you're right - just generate a truth image (ideally a fairly small one) and commit that along with the test.

A lot of the benchmark tests are slow already, so yours doesn't need to be particularly fast. If running all the iterations takes more than 10s and you can't speed it up by just reducing the test image size then #[ignore] it for now.

nicokoch:
I just pushed the proposed changes

theotherphil:
Great, thanks. I think you're missing the new image file. Worryingly Travis is green despite the test failing.

nicokoch:
Whoops, yeah. I just pushed it

PR #122, opened 2016-02-26
Fix canny algorithm artifacts.

nicokoch:
Previously, the canny algorithm produced artifacts, because the
non-maximum-suppression was done in-place.


theotherphil:
This mistake feels familiar! https://github.com/PistonDevelopers/imageproc/commit/c0dce1f8585a074494422683a9922fe1f152bffd

nicokoch:
Ehh, awkward :D

PR #123, opened 2016-03-03
Canny: implement hysteresis as breadth-first search

nicokoch:
This makes the algorithm a lot more correct, but is a little bit slower
than the previous approach (not noticeable at all).
I noticed this error while tinkering with parallel implementations,
where the processing order of the pixels was not reproducable.


theotherphil:
I like the sound of parallel implementations!

nicokoch:
I just posted an update on the relevant issue. Not much success yet, but I'm sure it will work one day :D

PR #133, opened 2016-04-25
Add adaptive thresholding algorithm.

nicokoch:
The algorithm traverses the image and calculates an adaptive threshold
on the fly, by taking the average brightness of the blacksize x
blocksize surrounding pixels.
This implementation uses integral images to reach O(n) performance.


nicokoch:
I just noticed the `block_size` argument is actually `block_size / 2`, let me fix this real quick before you review.

theotherphil:
Looks good generally, thanks! This seems like the sort of thing that's amenable to unit tests as well as regression tests (i.e. tests specifying exact input and output pixel values, as opposed to just checking that the results are plausible on a sample image). I've not checked the details too closely yet, but it would be nice to have tests showing that there are no off by one errors in the integral image points that are evaluated to get the mean.

theotherphil:
I'm on my phone right now so it's awkward to check, but I think everywhere else with a requirement that a side length is odd instead expresses the parameter as a radius that's then multiplied up. If we don't already have this convention then ignore me.

theotherphil:
I think this does have off by one issues with its use of integral images. e.g. If using a half_side of 1 on a 3*3 image, the mean for the centre pixel doesn't include the top left pixel. See eval_points in the Haar module for an example of dealing with this (although it's pretty clunky - you might have a neater solution).

nicokoch:
Yeah I definitely see the issues here. I actually coded this together quite some time ago, so my knowledge about integral images already faded a bit. I will read up again and put some fixes together

theotherphil:
Are you still intending to get this working? This is a nice function to have, so if you've lost interest I'll fix it up myself. No rush, just checking if this is dead or merely dormant!

nicokoch:
Sorry, I'm really busy jobwise atm, so I didn't really pay any more attention to this.

Sure, you can fix it up yourself :)

theotherphil:
Ok, thanks for the update. I'll fix this up soonish.

theotherphil:
Merged fixed implementation in 142 (and made integral_image less error prone - see comments there)

PR #135, opened 2016-06-17
Prevent drawing an extra line in draw_line_segment_mut

kevinmehall:
In cases where `is_steep` was `true`, (y, x) was out of bounds, but
(x, y) was in bounds, the `else` clause would draw an extra line segment by
mirroring the out-of-bounds part of the line into the image.

Example
```rust
extern crate image;
extern crate imageproc;

use image::{ImageBuffer, Rgb};

fn main() {
    let mut img = ImageBuffer::<Rgb<u8>, Vec<u8>>::new(400, 200);
    let red = Rgb { data: [255, 0, 0]};
    imageproc::drawing::draw_line_segment_mut(&mut img, (-100., 500.), (400., -300.), red);
    img.save("line.png").unwrap();
}
```

Before:
line-before

After:
line-after

### DIFF
--- a/src/drawing.rs
+++ b/src/drawing.rs
@@ -95,10 +95,14 @@ pub fn draw_line_segment_mut<I>(image: &mut I, start: (f32, f32), end: (f32, f32
 
     for x in x0 as i32..(x1 + 1f32) as i32 {
         unsafe {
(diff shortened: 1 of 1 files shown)


theotherphil:
Thanks!

PR #140, opened 2016-11-20
Initial implementation of Xu anti-aliasing

theotherphil:
Initial implementation of Xu's algorithm for anti-aliased lines. Does a bounds check and at least one unnecessary floating point operation per pixel, doesn't support the "include endpoint" parameter of https://github.com/wvanbergen/chunky_png, and only handles integer endpoints.

However, there's reasonable test coverage so it should be easy to improve all of these later.

(Towards https://github.com/PistonDevelopers/imageproc/issues/97)


korczis:
Wow. Great. Thanks for this.

theotherphil:
Let me know if you try it out and it doesn't do what you want :-)

PR #157, opened 2017-02-18
Added basic support for ttf and otf generic image font drawing

mhsjlw:
This is a work-in-progress, use this pull request for tracking checklist completion:

- [ ] handling of point sizes and converting between pixels and absolute sizes
- [x] positioning of font, and out-of-bounds checks
- [x] make it generic over pixel type
- [x] example
- [ ] newline support (?)
- [ ] outline support (?)
- [ ] add some regression tests
- [ ] add some unit tests - probably by writing the imageproc rendering function in terms of something with a signature like the draw function from rusttype, and then mocking out draw in the unit test ourselves


mhsjlw:
Note that currently this won't build due to:

```
error[E0277]: the trait bound `<<I as image::GenericImage>::Pixel as image::Pixel>::Subpixel: definitions::Clamp<f32>` is not satisfied
  --> src/drawing.rs:43:31
   |
43 |                   let color = weighted_sum(background, foreground, 1.0 - gv, gv);
   |                               ^^^^^^^^^^^^ the trait `definitions::Clamp<f32>` is not implemented for `<<I as image::GenericImage>::Pixel as image::Pixel>::Subpixel`
   |
   = help: consider adding a `where <<I as image::GenericImage>::Pixel as image::Pixel>::Subpixel: definitions::Clamp<f32>` bound
   = note: required by `pixelops::weighted_sum`

error[E0277]: the trait bound `f32: conv::ValueFrom<<<I as image::GenericImage>::Pixel as image::Pixel>::Subpixel>` is not satisfied
  --> src/drawing.rs:43:31
   |
43 |                   let color = weighted_sum(background, foreground, 1.0 - gv, gv);
   |                               ^^^^^^^^^^^^ the trait `conv::ValueFrom<<<I as image::GenericImage>::Pixel as image::Pixel>::Subpixel>` is not implemented for `f32`
   |
   = help: consider adding a `where f32: conv::ValueFrom<<<I as image::GenericImage>::Pixel as image::Pixel>::Subpixel>` bound
   = note: required because of the requirements on the impl of `conv::ValueInto<f32>` for `<<I as image::GenericImage>::Pixel as image::Pixel>::Subpixel`
   = note: required by `pixelops::weighted_sum`

```

If anyone can help me out with that, that'd be great. Thanks

examples/font.rs
```diff
@@ -0,0 +1,31 @@
...
+
+extern crate image;
+extern crate imageproc;
+extern crate rusttype;
+
+use std::path::Path;
+use imageproc::pixelops::weighted_sum;
+use image::{Rgb, RgbImage, open, ImageBuffer, GenericImage};
+use rusttype::{FontCollection, Scale, point, PositionedGlyph};
+
```
theotherphil: You don't need most of these use statements, right?

src/drawing.rs
```diff
@@ -8,6 +8,46 @@ use std::cmp::{min, max};
 use std::f32;
 use std::i32;
 
+use pixelops::weighted_sum;
+use rusttype::{FontCollection, Scale, point, PositionedGlyph};
+
+pub fn draw_text<I>(image: &mut I, color: I::Pixel, x: u32, y: u32, height: f32, offset: f32, scale: Scale, font_data: &[u8], text: &str) 
+    where I: GenericImage
+{
```
theotherphil: The compiler error you're seeing is because you're using weighted_sum, which requires Pixels satisfying `P::Subpixel: ValueInto<f32> + Clamp<f32>`, but this function is missing that constraint.

src/drawing.rs
```diff
@@ -8,6 +8,46 @@ use std::cmp::{min, max};
 use std::f32;
 use std::i32;
 
+use pixelops::weighted_sum;
+use rusttype::{FontCollection, Scale, point, PositionedGlyph};
+
+pub fn draw_text<I>(image: &mut I, color: I::Pixel, x: u32, y: u32, height: f32, offset: f32, scale: Scale, font_data: &[u8], text: &str)
```
theotherphil: This function needs a comment. In particular, the meaning of height, offset and scale isn't clear. It's also worth mentioning that newlines aren't supported, if that's still the case.

src/drawing.rs
```diff
@@ -8,6 +8,46 @@ use std::cmp::{min, max};
...
+    for g in glyphs {
+        if let Some(bb) = g.pixel_bounding_box() {
+            g.draw(|gx, gy, gv| {
+                let gx = gx as i32 + bb.min.x;
+                let gy = gy as i32 + bb.min.y;
+                
+                if gx >= 0 && gx < width as i32 && gy >= 0 && gy < pixel_height as i32 {
+
+                  let pixel = image.get_pixel(gx as u32 + x, gy as u32 + y);
+                  let foreground = color;
```
theotherphil: There's no need for the `foreground` variable.

src/drawing.rs
```diff
@@ -8,6 +8,46 @@ use std::cmp::{min, max};
...
+                  let background = pixel;
+                  let color = weighted_sum(background, foreground, 1.0 - gv, gv);
```
theotherphil: Or the `background` variable.

examples/font.rs
mhsjlw: I use all of these `use` statements, yes.

src/drawing.rs
mhsjlw: Totally missed that one, will do.

examples/font.rs
theotherphil: You use most of them in the implementation of draw_text, but not in this file (weighted_sum, for example, isn't used).

mhsjlw:
Here is example output from the `font.rs` example:

Example Image

theotherphil:
The following version handles out-of-bound pixels correctly. If you're happy with this version could you please amend the bounds checking in your pull request.

```
/// Draws colored text on an image in place. Note that this function *does not* support newlines, you must do this manually.
pub fn draw_text_mut<I>(image: &mut I, color: I::Pixel, x: u32, y: u32, scale: Scale, font_data: &[u8], text: &str)
    where I: GenericImage,
          <I::Pixel as Pixel>::Subpixel: ValueInto<f32> + Clamp<f32>
{
    let collection = FontCollection::from_bytes(font_data);
    let font = collection.into_font().expect("Unable to parse font data");

    let v_metrics = font.v_metrics(scale);
    let offset = point(0.0, v_metrics.ascent);

    let glyphs: Vec<PositionedGlyph> = font.layout(text, scale, offset).collect();

    for g in glyphs {
        if let Some(bb) = g.pixel_bounding_box() {
            g.draw(|gx, gy, gv| {
                let gx = gx as i32 + bb.min.x;
                let gy = gy as i32 + bb.min.y;

                let image_x = gx as u32 + x;
                let image_y = gy as u32 + y;

                if image_x >= 0 && image_x < image.width() && image_y >= 0 && image_y < image.height() {
                    let pixel = image.get_pixel(image_x, image_y);
                    let weighted_color = weighted_sum(pixel, color, 1.0 - gv, gv);
                    image.put_pixel(image_x, image_y, weighted_color);
                }
            })
        }
    }
}
```

mhsjlw:
@theotherphil Please take a look at my changes as well as a `draw_text` function that isn't mutable.

theotherphil:
Looks good, thanks! Just one tiny issue left before it's ready to merge - could you remove the reference to the `height` parameter from the doc comments.

mhsjlw:
Yup!

mhsjlw:
Do we want to explore at least some type of testing first?

theotherphil:
Sure! There are two sorts of tests we could write here: 

1. Basic regression tests for obvious edge cases
2. More fine-grained unit tests for correct handling of e.g. colour blending from coverage fraction

I think most the benefit will come from 1, for which we can add some tests to tests/regression.rs. So the existing drawing test there for examples. I think we should add a new test function which creates an empty image and then renders some text on it within bounds, some partially inside bounds and some totally outside bounds. It would also be good to test colours - maybe have an image which is one colour in the top half and another in the bottom half, and render some text in a third colour which partially covers both areas.

mhsjlw:
Yes, that sounds good. Would we just have pre-generated images that the output compares itself with?

theotherphil:
Yep! Have a look in tests/data for some other examples. Input images live in tests/data, and expected outputs in tests/data/truth. In this case we don't need to store the input image - we can easily generate it ourselves. regression.rs also contains some helper functions for updating results and doing the comparisons. The final function test_draw_filled_ellipse is a simple example of how to use them.

theotherphil:
@mhsjlw: rather than leaving this sitting open indefinitely I'll merge now and we can add any desired tests later. Thanks!

PR #178, opened 2017-05-17
Updated nalgebra to 0.11

johann2:
Nalgebra 0.5 doesn't compile with the latest nightly for some reason, so it needed update.
Depending on your feelings on nalgebra, it would be better to either use it everywhere in the library or to get rid of it for good.


theotherphil:
Thanks. We barely use it, so not using it at all would probably be best.

theotherphil:
The failing beta build stops me from merging on my phone, but I'll merge when I get home.

PR #188, opened 2017-06-07
Allow drawing of borrowed fonts

archer884:
I thought it would be nice to be able to use the same instance of `Font` for multiple calls to draw. If the library already has this feature, then I am sorry I missed it; I just thought it would be really handy to have, and I'd prefer to reference the canonical version of `imageproc` rather than my fork, so I thought I'd see if the project is interested.

My implementation here forwards all the other calls to a new function called `draw_text_with_font_mut()`, and a non-mut version of the same is provided. All the existing functions are still there; their implementation just moved to the new one.


theotherphil:
I don't think we need both versions, and I prefer the version taking a Font. Could you just change the signature of the existing function please. I don't think we need to worry much about backwards compatibility at this point.

archer884:
@theotherphil Absolutely! I'll get on that.

theotherphil:
Thanks! I think you still need to update examples/font.rs (and remove the `FontCollection` use? It doesn't look like this is necessary anymore) and I'll merge :-)

Issue #214, opened 2017-06-19
Remove trait bound requiring `HasBlack` for rotating images with `affine::rotate_with_default`


joek13:
As I understand, rotating images using `affine::rotate_with_default` shouldn't require the image's pixels to implement HasBlack. I understand how `rotate` would, as there's no understood default color. In my case, I'm trying to rotate an image with a transparent background (and therefore want the new background pixels to be transparent as well). 

The following code fails because `image::Rgba<u8>` doesn't implement `HasBlack`:

```
                transparent_image = rotate_with_default(&transparent_image,
                                                 (x, y),
                                                 3.1415,
                                                 Rgba([0, 0, 0, 0]),
                                                 Interpolation::Bilinear);
```
Why is this trait bound even necessary if the default color is going to be provided by the user anyways?

joek13:
I went ahead and forked the repo and removed the trait bound. Everything seems to be working fine. I didn't make a PR yet because it's literally a one-line change and thought you guys wouldn't want it. Let me know if its otherwise.

theotherphil:
Oops!. A pull request would be great, otherwise I'll fix this tomorrow.

theotherphil:
Fixed by #215

PR #215, opened 2017-06-20
remove HasBlack trait bound for rotate_with_default

joek13:
Addresses #214. Removes the trait bound requiring an image's `Pixel` type to implement `HasBlack` when rotating with `rotate_with_default` (which doesn't utilize the trait whatsoever).


theotherphil:
Thanks!

PR #248, opened 2018-01-29
Bump dependency Versions

Michael-F-Bryan:
This just updates various dependencies to their latest version. No changes were needed and all tests should still pass.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,13 +11,13 @@ exclude = ["examples/*.ttf"]
 [dependencies]
 conv = "0.3.1"
 image = "0.18.0"
-itertools = "0.6.0"
-nalgebra = "0.11"
+itertools = "0.7.0"
+nalgebra = "0.13"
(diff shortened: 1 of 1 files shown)


theotherphil:
Happy to merge, but is there a specific reason for this bump?

Michael-F-Bryan:
I'm working on a fairly large project at work and noticed that a lot of compilation time is wasted recompiling different versions of the same crate. A full release build will typically take 5-10 minutes, so I was hoping to decrease compilation times a bit by upgrading people's dependencies. 

In particular I believe `imageproc` currently uses older versions of `rayon` and `nalgebra`. With `nalgebra` pulling in compilation heavy things like `typenum`.

theotherphil:
Ah, right. Thanks.

theotherphil:
There's an open ticket for getting rid of the nalgebra dependency altogether, as we use almost nothing from it.

PR #264, opened 2018-05-18
Support of RgbaImage for affine::rotate

svalat:
I needed support for RGBA colors for `affine::rotate` in one of my project and it was missing. I made the patch. The only missing part was the implementation of HasBlack and HasWhite to get fallback color.

I also updated the test and add to ref images for the rotated elephan in RGBA. Fill free to not keep them if you want.

I'm not yet a Rust expert and was not able to run the test suite to check it pass, how you go throuth this :

```sh
error[E0554]: #![feature] may not be used on the beta release channel
 --> src/lib.rs:4:19
  |
4 | #![cfg_attr(test, feature(test))]
  |                   ^^^^^^^^^^^^^^

error: aborting due to previous error
```

I tried `cargo +beta test` or `cargo +nightly test` without any success. What you need to do to run your test suite ? (sorry for the stupid question).


tests/regression.rs
```diff
@@ -97,6 +97,14 @@ fn test_rotate_nearest_rgb() {
     compare_to_truth_rgb("elephant.png", "elephant_rotate_nearest.png", rotate_nearest_about_center);
 }
 
+#[test]
+fn test_rotate_nearest_rgba() {
+    fn rotate_nearest_about_center(image: &RgbaImage) -> RgbaImage {
+        rotate_about_center(image, std::f32::consts::PI/4f32, Interpolation::Nearest)
+    }
+    compare_to_truth_rgb("elephant_rgba.png", "elephant_rotate_nearest_rgba.png", rotate_nearest_about_center);
+}
```
theotherphil: This doesn't look like it would work - you're calling `compare_to_truth_rgb` with a function on `RgbaImage`s. I think you need to add a `compare_to_truth_rgba` method.

tests/regression.rs
```diff
@@ -111,6 +119,14 @@ fn test_rotate_bilinear_rgb() {
     compare_to_truth_rgb_with_tolerance("elephant.png", "elephant_rotate_bilinear.png", rotate_bilinear_about_center, 1);
 }
 
+#[test]
+fn test_rotate_bilinear_rgba() {
+    fn rotate_bilinear_about_center(image: &RgbImage) -> RgbImage {
```
theotherphil: Wrong type signature.

theotherphil:
Thanks for the PR! The tests only run on nightly, but `cargo +nightly test` should work. I don't think the new tests will compile though - I've left a couple of comments.

svalat:
Ok thanks for your nice eyes, I missed the `.to_rgb()` in `compare_to_truth_rgb_with_tolerance`. I made the rgba function.

Ok for running the tests, the issue is currently due to a bug which prevent to compile nalgebra-0.13.1 not imageproc.

theotherphil:
Thanks!

PR #265, opened 2018-05-28
Fix broken test for rgba image rotation

svalat:
OK, now all the libs build with nightly so I see my bugs and fix them. Rgba test rotation pass now. Sorry for the mistake in my previous pull request.


theotherphil:
Thanks!

PR #278, opened 2018-11-14
Update rusttype to 0.7

DoumanAsh:
While it might be technically a breaking change, `Font` and `PositionedGlyph` itself retain the same interface, so I think it should be ok to update it as there were numerous changes already

The only breaking change was `CacheBuilder` which is not used by imageproc, at least directly


theotherphil:
Thanks

Issue #280, opened 2018-11-24
Hough transform fails to detect lines under certain angles


kamirr:
I noticed that when a line points under a certain range of angles (between 90° and 135° when measured from the top of the x-axis), the imageproc's implementation of the Hough transform won't detect it.

To give an example, consider those two images:
![edges1](https://user-images.githubusercontent.com/7727537/48971294-dd267800-f017-11e8-8c22-e14e5c59c4bd.jpg)
![edges2](https://user-images.githubusercontent.com/7727537/48971293-dd267800-f017-11e8-9936-9b29b422b249.jpg)

The first one should result in the almost exact same set of lines as the second one, just with negated angles. Interestingly, that's not the case. Those are my results:

![res](https://user-images.githubusercontent.com/7727537/48971319-37273d80-f018-11e8-8d93-d8d4e919f303.png)
![res](https://user-images.githubusercontent.com/7727537/48971322-3db5b500-f018-11e8-8c35-d8db5794302f.png)

Code:
```extern crate imageproc;
extern crate image;

fn main() {
    let args: Vec<String> = std::env::args().collect();
    assert_eq!(args.len(), 2, "Usage: <progname> <edges image>");

    let edges = match image::open(&args[1]).unwrap() {
        image::DynamicImage::ImageRgb8(img) => img,
        _ => panic!("Passed file should be RGB")
    };

    let grayscale = image::GrayImage::from_fn(edges.width(), edges.height(), |x, y| {
        let px = edges.get_pixel(x, y);
        let r = px.data[0] as f64;
        let g = px.data[1] as f64;
        let b = px.data[2] as f64;

        image::Luma{ data: [(0.299 * r*r + 0.587 * g*g + 0.114 * b*b).sqrt() as u8] }
    });

    let vote_threshold = 180;
    let lines = imageproc::hough::detect_lines(&grayscale, imageproc::hough::LineDetectionOptions {
        vote_threshold,
        suppression_radius: 0
    });

    imageproc::hough::draw_polar_lines(
        &edges,
        &lines,
        image::Rgb {
            data: [255, 0, 255]
        }
    ).save("res.png").unwrap();
}
```

kamirr:
I've looked into the implementation of `detect_lines` and managed to get it to work:
```
/// Detects lines in a binary input image using the Hough transform.
///
/// Points are considered to be in the foreground (and thus vote for lines)
/// if their intensity is non-zero.
///
/// See ./examples/hough.rs for example usage.
[...]

PR #282, opened 2018-11-25
Hough fixes

theotherphil:
WIP fix for https://github.com/PistonDevelopers/imageproc/issues/280


src/hough.rs
```diff
@@ -55,14 +55,11 @@ pub fn detect_lines(image: &GrayImage, options: LineDetectionOptions) -> Vec<Pol
             let p = unsafe { image.unsafe_get_pixel(x, y)[0] };
 
             if p > 0 {
-                for m in 0..180u32 {
+                for (m, (c, s)) in cos_lut.iter().zip(&sin_lut).enumerate() {
```
HeroicKatora: It could be faster and clearer to simply merge the two vectors into one vector `Vec<(f32, f32)>` now that I think about it.

src/hough.rs
```diff
@@ -42,7 +43,9 @@ pub fn detect_lines(image: &GrayImage, options: LineDetectionOptions) -> Vec<Pol
     let rmax = ((width * width + height * height) as f64).sqrt() as i32;
 
     // Measure angles in degrees, and use bins of width 1 pixel and height 1 degree.
-    let mut acc: ImageBuffer<Luma<u32>, Vec<u32>> = ImageBuffer::new(rmax as u32, 180u32);
+    // We use the convention that distances are positive for angles in (0, 180] and
+    // negative for angles in [180, 360).
+    let mut acc: ImageBuffer<Luma<u32>, Vec<u32>> = ImageBuffer::new(2 * rmax as u32 + 1, 180u32);
```
HeroicKatora: This introduces an index error in the unsafe below, I think. If `rmax = 1.6` then:

`2 + rmax as u32 + 1 = 3`

but at the same time, we can get

`r=1.4 => d as u32 = (r + rmax) as u32 = (1.4 + rmax) as u32 = 3`
`d = 3 <= 2*rmax = 3.2`

Edit: nvm, rmax is `i32`, I forgot cast happens before.

HeroicKatora: Sorry.

src/hough.rs
```diff
@@ -81,7 +82,7 @@ pub fn detect_lines(image: &GrayImage, options: LineDetectionOptions) -> Vec<Pol
             let votes = unsafe { acc_sup.unsafe_get_pixel(r, m)[0] };
             if votes >= options.vote_threshold {
                 let line = PolarLine {
-                    r: r as f32,
+                    r: r as f32 - rmax as f32,
```
HeroicKatora: For large images (diagonal longer than `1 << 23`), the cast and subtraction are inaccurate in `f32` but at least the subtraction would be accurate in `u32/i32`. Maybe rewrite as `(r - rmax as u32) as i32 as f32`. 

The inaccuracy affects all lines for larger images. Especially those detected with an offset close to `0` have a high relative error. After the change, only affectes lines with an offset greater than `1 << 23`.

src/hough.rs
```diff
@@ -133,7 +134,8 @@ fn intersection_points(
     image_width: u32,
     image_height: u32,
 ) -> Option<((f32, f32), (f32, f32))> {
-    let r = line.r;
+    // FIXME: handle this properly
+    let r = if line.r == 0.0 { line.r + 0.0001 } else { line.r };
```
HeroicKatora: This apparently wants to avoid the cases of division complications and instabilities?

Turning the division into multiplications of the other comparison side, correctly keeping in mind the comparison operator change for `sin(theta) < 0`, might solve this issue. This would effectively simulate the necessary computations in homogeneous coordinates and be more stable.

So for example:

```
let left_y = r * sin.signum();
…
if left_y >= 0 && left_y < h*sin.abs() {
}
```

HeroicKatora: For `(r - w * cos)` itself, `cos.mul_add(-w, r)` could be faster and more accurate as well.

Especially important where `sin` is close to 0, but `w*cos` is also close to `r`, i.e. the line is nearly vertical and near the outermost side of the image.

src/hough.rs
```diff
@@ -248,6 +257,84 @@ mod test {
...
+                assert_approx_eq!(line.r, $r, 1.1);
+                assert_approx_eq!(line.angle_in_degrees as f32, $angle as f32, 5.0);
+            }
+        };
+    }
+
+    test_detect_line!(detect_line_50_45, 50.0, 45);
+    test_detect_line!(detect_line_eps_135, 0.001, 135);
+    // https://github.com/PistonDevelopers/imageproc/issues/280
+    test_detect_line!(detect_line_neg10_120, -10.0, 120);
```
HeroicKatora: Those look like a good fit for a benchmark test as well, finishing the TODO below.

src/hough.rs
theotherphil: I still need to address this comment. I'm going to rewrite this function and add some tests.

theotherphil: It's not right; it was just the quickest hack to test that I'd fixed the bug reported in #280. I've just pushed a commit that removes this line.

theotherphil:
@HeroicKatora I think I've now addressed all your comments

PR #289, opened 2018-11-27
Document limitation of connected_components and panic early

theotherphil:
Fixes https://github.com/PistonDevelopers/imageproc/issues/285

@HeroicKatora I've gone for the easiest (laziest) option for now; I won't bother rewriting the function to handle larger images until I know there's anyone that actually needs this.


HeroicKatora:
Seems like a valid solution until someone encounters the commend :+1:

PR #297, opened 2018-12-19
Make quickcheck and assert_approx_eq dev dependencies

selaux:
Saves any dependants a little compile time, as those crates are only used for testing as far as I can see.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,6 @@ homepage = "https://github.com/PistonDevelopers/imageproc"
 exclude = ["examples/*.ttf"]
 
 [dependencies]
-assert_approx_eq = "1.0.0"
 conv = "0.3.1"
 image = "0.20.0"
 itertools = "0.7.0"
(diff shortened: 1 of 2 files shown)


theotherphil:
Thanks for the PR. Does this change still allow other crates to use the property_testing module for their own tests?

selaux:
It does not. I though the property_testing module was there for property testing utilities. Would it make sense to make that a feature? I would also not mind to close the PR :wink:, it saves only very little of compile time anyways...

theotherphil:
I think I've seen someone using this library specifically for the testing utils, but can't remember where. In any case, it seems like a potentially useful thing to expose, so making it available behind a feature flag sounds ideal.

selaux:
Allright I hid the module behind a feature flag. I made it opt-in, as my guess was that most consumers won't use property testing. Please change if you think it is appropriate.

Thanks for the library by the way... its very useful.

theotherphil:
Looks good, thanks.

Glad to hear someone is using it!

Issue #277, opened 2018-11-13
Add the ability to output image to a window using some graphics api and probably winit


lazypassion:
Use a graphics api like vulkano or glium to output images to a window(probably using winit) so you can see the output of the different processing algorithms. It would probably help with informal testing of algorithm correctness and faster iteration when writing or using them.

theotherphil:
I agree that this would be useful. I don't want to pull in extra graphics dependencies unconditionally so this should be enabled via a package feature.

I've not felt any need to add this myself because I've not got round to finding a decent debugger setup and so just write files to disk instead. Got any tips on debuggers for Rust?

lazypassion:
I'm also in the same boat. I've tried debugging on vscode and it isn't as satisfactory as the ones found in ides from jetbrains. I've heard clion might be good but I don't have the money to invest into it.

theotherphil:
Thanks for the tip - I'll give the codelldb extension a go.

lazypassion:
Would winit be preferred for the output window or is it ok to use sdl2, although it requires a c library

theotherphil:
My main constraint is that this functionality is hidden behind a feature flag. Beyond that, I'm happy for it to be provided in any sensible form, even if it requires (clearly documented!) extra dependencies or installed libraries.

theotherphil:
That would be cool to have, but it wouldn't totally remove the need for this feature - it's useful to be able to debug an image processing operation somewhere in the middle of a larger piece of code.

theotherphil:
I've seen this example in the image-png repo, which might be a useful starting point:

https://github.com/PistonDevelopers/image-png/blob/master/examples/show.rs

PR #298, opened 2019-01-08
Adds ability to display an image in a window using sdl2 

lazypassion:
This is for #277.

This is my first commit to any repo that isn't my own, so criticism is welcome!

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -121,6 +121,7 @@ pub fn display_image(title: &str, image: &RgbaImage, window_width: u32, window_h
                     let center_y = ((y - height) as f32 / 2.0_f32) as i32; 
                     canvas.clear(); 
                     canvas.copy(&texture, None, Rect::new(center_x, center_y, width, height)).unwrap();
+                    canvas.present();
                 },           
                 _ => {}
             }
--- a/src/window.rs
+++ b/src/window.rs
@@ -12,7 +12,7 @@ pub fn display_image(title: &str, image: &RgbaImage, window_width: u32, window_h
     const MIN_WINDOW_DIMENSION: u32 = 150;
     // ensures window size is minimum size, so that image resizing calculations for the window are correct
     let window_width: u32 = if window_width < MIN_WINDOW_DIMENSION {
-        MIN_WINDOW_DIMENSION 
+        MIN_WINDOW_DIMENSION
     } else {
         window_width
     };
@@ -23,7 +23,11 @@ pub fn display_image(title: &str, image: &RgbaImage, window_width: u32, window_h
     };
 
     // resizes and returns the image that will be used to display in the window
-    fn create_display_image(image: &RgbaImage, window_width: u32, window_height: u32) -> (u32, u32, RgbaImage) {
+    fn create_display_image(
+        image: &RgbaImage,
+        window_width: u32,
+        window_height: u32,
+    ) -> (u32, u32, RgbaImage) {
         if image.height() < window_height && image.width() < window_width {
             (image.width(), image.height(), image.clone())
         } else {
@@ -45,8 +49,9 @@ pub fn display_image(title: &str, image: &RgbaImage, window_width: u32, window_h
             (width, height, output_image)
         }
     }
-    
-    let (output_image_width, output_image_height, output_image) = create_display_image(image, window_width, window_height);
+
+    let (output_image_width, output_image_height, output_image) =
+        create_display_image(image, window_width, window_height);
 
     const CHANNEL_COUNT: u32 = 4;
     let pitch = output_image_width * CHANNEL_COUNT;
@@ -119,7 +124,8 @@ pub fn display_image(title: &str, image: &RgbaImage, window_width: u32, window_h
                     let x = x as u32;
                     let y = y as u32;
                     // resize image if necessary to fit into the window
-                    let (output_image_width, output_image_height, output_image) = create_display_image(image, x, y);
+                    let (output_image_width, output_image_height, output_image) =
+                        create_display_image(image, x, y);
 
                     let pitch = output_image_width * CHANNEL_COUNT;
                     let mut img_raw = output_image.into_raw();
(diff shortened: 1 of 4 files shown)


lazypassion:
On a quick check it looks like that is possible, so I'll look into adding it. And btw, the performance of rendering on the window is pretty abysmal with optimization levels 0 and 1(basically debug mode).

lazypassion:
I would prefer to add the ability to create multiple image windows in another pull request. Having a way to display a single image is sufficient for some use cases and multiple windows satisfies another use case.

src/window.rs
```diff
@@ -0,0 +1,149 @@
...
+
+use image::RgbaImage;
+use sdl2::event::{WindowEvent, Event};
+use sdl2::rect::Rect;
+use sdl2::keyboard::Keycode;
+use sdl2::pixels::{Color, PixelFormatEnum};
+use sdl2::surface::Surface;
+
+/// Displays the provided RGBA image in a new window.
+pub fn display_image(title: &str, image: &RgbaImage, window_width: u32, window_height: u32) {
```
theotherphil: It would be nice if this was generic over pixel type, but I'm happy to merge without that.

Please mention the min window size here so that no-one gets confused when they try to draw a 100x100 window and it doesn't work.

src/window.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    let image_width = image.width();
```
theotherphil: You could instead write `let (image_width, image_height) = image.dimensions()` if you wanted to save a whole one line.

src/window.rs
```diff
@@ -0,0 +1,149 @@
...
+    let image_height = image.height();
+
+    // resizes and returns the image that will be used to display in the window
+    let (output_image_width, output_image_height, output_image) = 
+        if image_height < window_height && image_width < window_width {
+            (image_width, image_height, image.clone())
+        } else {
+            // scale is used to determine how small an image has to be resized to fit within
+            // the provided window dimensions
+            let scale = {
```
theotherphil: This is wrong if either window_width or window_height is less than 300. You need to define the min window width and height earlier and adjust the user-provided width and height by the time you do this scaling.

src/window.rs
```diff
@@ -0,0 +1,149 @@
...
+                let width_scale = window_width as f32 / image_width as f32; 
+                let height_scale = window_height as f32 / image_height as f32;
+                if  width_scale < height_scale {
```
theotherphil: Nit: two spaces after "if"

src/window.rs
```diff
@@ -0,0 +1,149 @@
...
+                }
+            };
+            let height = (scale * image_height as f32) as u32;
+            let width = (scale * image_width as f32) as u32;
+            let output_image = image::imageops::resize(image, width, height, image::FilterType::Triangle);
+            (width, height, output_image)
+    };
+
+    const CHANNEL_COUNT: u32 = 4;
+    let pitch  = output_image_width * CHANNEL_COUNT;
```
theotherphil: Nit: two spaces after "pitch" - worth running rustfmt and using whatever formatting that creates?

src/window.rs
```diff
@@ -0,0 +1,149 @@
...
+                Event::Quit {..} |
+                    Event::KeyDown {keycode: Some(Keycode::Escape), .. } => {
+                        break 'running
+                    },
+                Event::Window { win_event: WindowEvent::Resized(x, y), .. } => {
+                    
+                    let x = x as u32;
+                    let y = y as u32;
+                    // resize image if necessary to fit into the window
+                    let (output_image_width, output_image_height, output_image) =
```
theotherphil: This looks suspiciously similar to the scaling code above - could you extract the scaling code into a function private function and use it in both places?

src/window.rs
```diff
@@ -0,0 +1,149 @@
...
+                            let output_image = image::imageops::resize(
+                                image, 
+                                width, 
+                                height, 
+                                image::FilterType::Triangle);
+                            (width, height, output_image)
+                        };
+                    
+
+                    let pitch  = output_image_width * CHANNEL_COUNT;
```
theotherphil: This next section also appears to be copy-pasted from above, so should be moved into function to be called from both locations.

src/window.rs
lazypassion: I wanted to make the function generic too, however, the function becomes a lot more complex, because inevitably the input image has to be converted to an RGB image. I figured it would be easier and simpler for the user to call convert from the ConvertBuffer trait instead of having to figure out what kind of image is being passed into the function and correctly converting it to RGB.

src/window.rs
lazypassion: Would it be fine to make it a nested function?

theotherphil: Sure

lazypassion:
I'll squash all the commits if everything looks good.

src/window.rs
```diff
@@ -0,0 +1,162 @@
...
+use sdl2::pixels::{Color, PixelFormatEnum};
+use sdl2::rect::Rect;
+use sdl2::surface::Surface;
+
+/// Displays the provided RGBA image in a new window.
+/// Minimum window size is 150 x 150.
+pub fn display_image(title: &str, image: &RgbaImage, window_width: u32, window_height: u32) {
+    const MIN_WINDOW_DIMENSION: u32 = 150;
+    // ensures window size is minimum size, so that image resizing calculations for the window are correct
+    let window_width: u32 = if window_width < MIN_WINDOW_DIMENSION {
```
theotherphil: u32 implements Ord, so this can just be `let window_width = window_width.max(MIN_WINDOW_DIMENSION)`

src/window.rs
```diff
@@ -0,0 +1,162 @@
...
+        MIN_WINDOW_DIMENSION
+    } else {
+        window_width
+    };
+    let window_height: u32 = if window_height < MIN_WINDOW_DIMENSION {
```
theotherphil: Same here

theotherphil:
Fixes https://github.com/PistonDevelopers/imageproc/issues/277

PR #303, opened 2019-01-17
Enforce non-zero channel count in BresenhamLinePixelIterMut to prevent UB

theotherphil:
Fixes https://github.com/PistonDevelopers/imageproc/issues/281

cc @HeroicKatora

### DIFF
--- a/src/drawing/line.rs
+++ b/src/drawing/line.rs
@@ -14,7 +14,7 @@ pub struct BresenhamLineIter {
     error: f32,
     end_x: i32,
     is_steep: bool,
-    y_step: f32,
+    y_step: i32,
 }
 
 impl BresenhamLineIter {
@@ -45,7 +45,7 @@ impl BresenhamLineIter {
             error: dx / 2f32,
             end_x: x1 as i32,
             is_steep: is_steep,
-            y_step: if y0 < y1 { 1f32 } else { -1f32 }
+            y_step: if y0 < y1 { 1 } else { -1 }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


HeroicKatora:
This is the basic right thing. I'm also a bit worried about the number of casts and type conversion that go on during the evaluation. For example:

1. Suppose `end.0 == i32::max_value()`. Then https://github.com/PistonDevelopers/imageproc/blob/143350310fb5ec0abe1865050ac59f864c2dc992/src/drawing/line.rs#L57 can never be true.
2. Therefore, https://github.com/PistonDevelopers/imageproc/blob/143350310fb5ec0abe1865050ac59f864c2dc992/src/drawing/line.rs#L66 will silently overflow in 2's complement in release. 
3. Then this negative `i32` is again silently converted to an exceedingly positive index at https://github.com/PistonDevelopers/imageproc/blob/143350310fb5ec0abe1865050ac59f864c2dc992/src/drawing/line.rs#L146
4. If `image::width() == u32::max_value()` this can wrap again to `0` without ever triggering the out-of-bounds from the pixel access.

HeroicKatora:
Btw, I see no real reason why `y_step` is `f32`. It is only assigned `1` or `-1` and only used to add to a `i32`.

theotherphil:
Thanks, I've made those changes too.

theotherphil:
Actually, only the channel count check is needed - BresenhamLinePixelIterMut clamps the input coordinates to the image dimensions anyway.

theotherphil:
Oops, I also need to compare against i32::max_value when clamping. Sigh. Will do that in a future commit.

HeroicKatora:
Yeah of course, the check of ` < i32::max_value() ` can happen at construction time.

PR #307, opened 2019-01-20
Display multiple images implementation, some minor fixes

### LINKED ISSUE #301: Displaying multiple images
Should displaying multiple images happen in multiple windows or in one window?

lazypassion:
#301


theotherphil:
Thanks! I’ll have a look through this tomorrow.

lazypassion:
I was getting segmentation fault's sometimes when running it on ubuntu 18.04, so tell me if you experience any issues. It seems to be a problem with sdl2 because it works fine on windows.

lazypassion:
Yeah it looks like it can, I'll make the change if nothing weird crops up

src/window.rs
```diff
@@ -23,53 +47,93 @@ pub fn display_image(title: &str, image: &RgbaImage, window_width: u32, window_h
...
+    let mut windows: Vec<sdl2::video::Window> = Vec::with_capacity(images.len());
+    let mut window_visibility: Vec<bool> = Vec::with_capacity(images.len());
+    for _ in 0..images.len() {
+        let mut window = video_subsystem
+            .window(title, window_width, window_height)
+            .position_centered()
+            .resizable()
+            .build()
+            .expect("couldn't create window");
+
```
theotherphil: Creating all of the images in the same location is a bit confusing. You could use `set_position` to stagger each window by some multiple of its index.

lazypassion: I have them staggered, however there is an issue if there are too many windows it is possible that they are rendered off screen. So i'll have to add some code to prevent that. And by too many windows, I mean around 20 and more.

lazypassion:
If and when you merge, can you squash and merge please?

lazypassion:
@theotherphil is this how u ping?

theotherphil:
Thanks!

PR #311, opened 2019-02-01
Add Cross-Correlation to Template Matching

JonBoyleCoding:
I added Cross-Correlation as a method for template matching based on the equations here: https://docs.opencv.org/3.4/de/da9/tutorial_template_matching.html


theotherphil:
Looks good, thanks. It would be nice to have a new test for this function, and to add a comment explaining that (unlike SSE) higher values are better.

JonBoyleCoding:
Ok, I'll work on getting a test in!

JonBoyleCoding:
I've added tests based on the ones that are currently there, and I checked the expected result with OpenCV 4 to make sure it was correct.

theotherphil:
Thanks!

PR #319, opened 2019-04-12
Add  non-square kernel support to median filter

TH3CHARLie:
I add non-square kernel support for the median filter. It now supports any rectangle shape kernels with odd side lengths.

### DIFF
--- a/src/filter/median.rs
+++ b/src/filter/median.rs
@@ -2,7 +2,7 @@ use image::{GenericImageView, Pixel};
 use crate::definitions::Image;
 use std::cmp::{min, max};
 
-/// Applies a median filter of given kernel to an image. Each output pixel is the median
+/// Applies a median filter of given dimensions to an image. Each output pixel is the median
 /// of the pixels in a `(2 * x_radius + 1) * (2 * y_radius + 1)` kernel of pixels in the input image.
 ///
 /// Pads by continuity. Performs O(max(x_radius, y_radius)) operations per pixel.
@@ -83,7 +83,6 @@ use std::cmp::{min, max};
 /// # fn main() {
 /// use imageproc::filter::median_filter;
 ///
-/// // The kernel can be non-square shape.
 /// // This example uses a kernel with x_radius sets to 2
 /// // and y_radius sets to 1, which leads to 5 * 3 kernel size.
 ///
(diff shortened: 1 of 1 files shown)


src/filter/median.rs
```diff
@@ -2,10 +2,10 @@ use image::{GenericImageView, Pixel};
 use crate::definitions::Image;
 use std::cmp::{min, max};
 
-/// Applies a median filter of given `radius` to an image. Each output pixel is the median
-/// of the pixels in a `2 * radius + 1` square of pixels in the input image.
+/// Applies a median filter of given kernel to an image. Each output pixel is the median
```
theotherphil: "kernel" is wrong here. "of given dimensions" would be ok.

src/filter/median.rs
```diff
@@ -72,10 +72,41 @@ use std::cmp::{min, max};
...
-pub fn median_filter<P>(image: &Image<P>, radius: u32) -> Image<P>
+/// 
+/// ```
+/// # extern crate image;
+/// # #[macro_use]
+/// # extern crate imageproc;
+/// # fn main() {
+/// use imageproc::filter::median_filter;
+///
+/// // The kernel can be non-square shape.
```
theotherphil: This line is not necessary

src/filter/median.rs
```diff
@@ -289,43 +322,50 @@ mod tests {
...
+    bench_median_filter!(bench_median_filter_s100_r8, side: 100, x_radius: 8,y_radius: 8);
+    
+    // benchmark on non-square kernels
+    bench_median_filter!(bench_median_filter_s100_rx1_ry4, side: 100, x_radius: 1,y_radius: 4);
+    bench_median_filter!(bench_median_filter_s100_rx1_ry8, side: 100, x_radius: 1,y_radius: 8);
+    bench_median_filter!(bench_median_filter_s100_rx4_ry8, side: 100, x_radius: 4,y_radius: 1);
+    bench_median_filter!(bench_median_filter_s100_rx8_ry1, side: 100, x_radius: 8,y_radius: 1);
+
 
-    // Reference implementation of median filter - written to be as simple as possible,
```
theotherphil: Why did you remove this comment?

src/filter/median.rs
```diff
@@ -348,20 +388,21 @@ mod tests {
     fn median(sorted: &[u8]) -> u8 {
         let mid = sorted.len() / 2;
         sorted[mid]
-    }
+    }
```
theotherphil: Bad whitespace

src/filter/median.rs
TH3CHARLie: Unintentionally deleted it, recover it in the new commit

Issue #261, opened 2018-04-07
Can't draw semi-transparent shapes


mark-summerfield:
I want to be able to draw semi-transparent rectangles. I have tried doing this using colors with a suitable alpha channel. Take this code for example:
```rust
extern crate image;
extern crate imageproc;

use std::path::Path;
use image::{Rgba, RgbaImage};
use imageproc::drawing::draw_filled_rect_mut;

fn main() {
    let path = Path::new("1.png");
    let solid_white = Rgba([255u8, 255u8, 255u8, 255u8]);
    let blue = Rgba([0u8, 0u8, 255u8, 255u8]);
    let semi_red = Rgba([255u8, 0u8, 0u8, 127u8]);
    let mut img = RgbaImage::new(200, 200);
    draw_filled_rect_mut( // Solid white background
        &mut img, imageproc::rect::Rect::at(0, 0).of_size(200, 200),
        solid_white);
    draw_filled_rect_mut( // Solid blue square
        &mut img, imageproc::rect::Rect::at(25, 25).of_size(100, 100),
        blue);
    draw_filled_rect_mut( // Semi-transparent red square
        &mut img, imageproc::rect::Rect::at(75, 75).of_size(75, 75),
        semi_red);
    img.save(path).unwrap();
}
```
I expected this to draw a 200x200 square in solid white (which it does), then to draw a 100x100 solid blue square on top (which it does), then to draw a 75x75 semi-transparent red square that overlaps the blue and the white (which it does -- but not quite right). I expected the red to "show through" some blue where it overlaps the blue square and to "show through" some white (i.e., look pink) where it overlaps the white background.

But there is no show through at all.

Is it possible to achieve this effect? If so, where is it documented?

HeroicKatora:
You could draw the transparent shape into a new image, then blend the two images together. While that should have the desired effect, it's memory and performance is suboptimal. The issue with the current approach is that internally the library just sets the color of all pixels when executing `draw_filled_rect_mut`.

I think your expectation is correct and the interface should allow specifying a blend mode, but at least make it clearer what happens.

theotherphil:
This is not currently supported, but the Pixel trait has a [blend method](https://docs.rs/image/0.18.0/image/trait.Pixel.html#tymethod.blend) already so it should be easy to add.

Note however that I'm the main contributor to this library and won't be working on this any time soon. PRs are obviously welcome.

mark-summerfield:
[...]

PR #326, opened 2019-07-02
Introduce Canvas trait, make most mutating drawing functions generic over it

theotherphil:
Fixes https://github.com/image-rs/imageproc/issues/261

I'm not totally convinced by this API - it means that choosing a blending function requires the ability to take ownership of an image. However, it increases flexibility at no cost (except for a slightly less simple API).


HeroicKatora:
> I'm not totally convinced by this API - it means that choosing a blending function requires the ability to take ownership of an image.

Not necessarily. At worst, one can use [`image.sub_image(0, 0, width, height)`](https://docs.rs/image/0.21.2/image/trait.GenericImage.html#method.sub_image) to use a mutable borrow of the original image. Not sure how the optimizer likes that but its possible. I wouldn't question the current design, only whether `image` should have `impl GenericImage for &'_ mut GenericImage` similar to `std::io::{Read, Write}`.

Overall, this looks like a decent idea and execution. I'll do a more thorough review in the coming days.

theotherphil:
@HeroicKatora there's no rush, but do let me know if you don't think you'll have the time to review this and I'll just merge it without a review.

src/utils.rs
```diff
@@ -216,6 +216,108 @@ macro_rules! rgb_image {
...
+            use image::{ImageBuffer, Rgba};
+            ImageBuffer::<Rgba<$channel_type>, Vec<$channel_type>>::new(0, 0)
+        }
+    };
+    // Non-empty image of default channel type u8
+    ($( $( [$r: expr, $g: expr, $b: expr, $a:expr]),*);*) => {
+        rgba_image!(type: u8, $( $( [$r, $g, $b, $a]),*);*)
+    };
+    // Non-empty image of given channel type
+    (type: $channel_type:ty, $( $( [$r: expr, $g: expr, $b: expr, $a: expr]),*);*) => {
```
HeroicKatora: The explicit type argument in this manner seems rather surprising for Rust. Can it not be deducted?

src/drawing/canvas.rs
```diff
@@ -0,0 +1,65 @@
...
+/// A surface for drawing on - many drawing functions in this
+/// library are generic over a `Canvas` to allow the user to
+/// configure e.g. whether to use blending.
+///
+/// All instances of `GenericImage` implement `Canvas`, with
+/// the behaviour of `draw_pixel` being equivalent to calling
+/// `set_pixel` with the same arguments.
+pub trait Canvas {
+    /// The type of `Pixel` that can be drawn on this canvas.
+    type Pixel;
```
HeroicKatora: Is it intentional that this does not require `Pixel: image::Pixel`? The trait bound seems mentioned everywhere `Canvas` as used, so introducing it here would allow omitting it. This happens with the current `GenericImage` trait bound.

src/drawing/canvas.rs
```diff
@@ -0,0 +1,65 @@
...
+    Pixel
+};
+
+/// A surface for drawing on - many drawing functions in this
+/// library are generic over a `Canvas` to allow the user to
+/// configure e.g. whether to use blending.
+///
+/// All instances of `GenericImage` implement `Canvas`, with
+/// the behaviour of `draw_pixel` being equivalent to calling
+/// `set_pixel` with the same arguments.
```
HeroicKatora: This could use an example of using a `GenericImage` to show that it really is drop-in. 

It would be good to have a mention of `Blend` in the documentation as well, as motiviation for this intermediate trait and as an implementation guide for other `Canvas` customizations.

src/utils.rs
theotherphil: I remember having issues with type deduction when writing the `rgb_image` macro that this is based on, but I don't remember what they were. Most of the macros in this library use type `u8`, so hit the rule above this one and don't need to specify a type. I won't attempt to resolve the issue as part of this PR.

src/drawing/canvas.rs
theotherphil: There's no good reason for the current behaviour. I've pushed a commit to adopt your suggestion.

src/drawing/canvas.rs
theotherphil: 👍 Done

PR #329, opened 2019-07-04
Circle drawing functions are incorrect, so forward them to the ellipse drawing functions instead

theotherphil:
Fixes https://github.com/image-rs/imageproc/issues/328

PR #330, opened 2019-07-04
Fix circle drawing

### LINKED ISSUE #328: `draw_filled_circle_mut` gives an ugly circle when radius is small
This happens in when I try to draw a circle with small radius.

## Expected

(draw by pillow)
circle_py

## Actual behaviour

circle

(a more intuitive picture)
图片

## Reproduction steps

```rust
let mut circle = DynamicImage::new_luma8(21, 21);

draw_filled_circle_mut(
    &mut circle,
    (10, 10),
    10,
    Rgba([255, 255, 255, 255]),
);
circle.save(std::path::Path::new("circle.png"));
```

```python
circle = Image.new('L', (21, 21), 0)
draw = ImageDraw.Draw(circle)
draw.ellipse((0, 0, 20, 20), fill=255)
circle.save('circle_py.png')
```

ksqsf:
Fixes #328


theotherphil:
Argh, race condition! I've just opened https://github.com/image-rs/imageproc/pull/329

Presumably your PR also needs to update the expected output for test_draw_hollow_ellipse and test_draw_filled_ellipse.

HeroicKatora:
@ksqsf I'm very confused about the changes. The previous version was a well known version to track the circle entirely in integer arithmetic but this one I don't know.

ksqsf:
fighting with git... wait a moment :-(

src/drawing/conics.rs
```diff
@@ -197,11 +197,11 @@ where
 {
     let mut x = radius;
     let mut y = 0i32;
-    let mut err = 0i32;
+    let mut p = 1 - radius;
     let x0 = center.0;
     let y0 = center.1;
 
-    while x >= y {
+    while x <= y {
```
theotherphil: The initial value of `x` is larger than the initial value of `y` so it looks like this loop will never get executed.

ksqsf:
@HeroicKatora This is a tested textbook algorithm, originally written in C. Hopefully I didn't get anything wrong in porting it to Rust...

PS: the short name `p` means the decision parameter. (Maybe I should have given it a better name.)

src/drawing/conics.rs
ksqsf: Fixed in the new commit.

HeroicKatora:
@ksqsf The current one also looks like the textbook version I remember. The current version is tracking the error from square radius and checks if the error can be decreased by going a pixel down, your textbook seems to track the *relative* error from going a pixel down compared to current error but otherwise they are identical (except some integer overflows).

Your textbook version could be slightly faster (less arithmetic in the `if`-branch comparison but I doubt it makes a big difference, and it certainly does not fix whatever is the cause of the current bug.

ksqsf:
It gives better results when the radius is small though

image

HeroicKatora:
What's the result with the current?

ksqsf:
Updated version:
new

Current version:
old

Note the filled circle on the right.

ksqsf:
OK so I decreased the radius of the hollow circles as well.

new

old

theotherphil:
I'm not totally clear on how this fixes the issue, but it does look to fix it! I'll merge this one in favour of my PR. I've got a few small style nits.

src/drawing/conics.rs
```diff
@@ -211,11 +211,12 @@ where
         draw_if_in_bounds(image, x0 + y, y0 - x, color);
         draw_if_in_bounds(image, x0 + x, y0 - y, color);
 
-        y += 1;
-        err += 1 + 2 * y;
-        if 2 * (err - x) + 1 > 0 {
-            x -= 1;
-            err += 1 - 2 * x;
+        x += 1;
+        if (p < 0) {
```
theotherphil: Nits: remove the parens around `p < 0`.

src/drawing/conics.rs
```diff
@@ -211,11 +211,12 @@ where
...
+            p += 2*x + 1;
```
theotherphil: Nit: add a space on both sides of all arithmetic operators here and below.

src/drawing/conics.rs
ksqsf: Fixed.

src/drawing/conics.rs
ksqsf: Fixed.

HeroicKatora:
In other words, the current implementation is simply wrong? Well great ^^ Thanks a ton @ksqsf

theotherphil:
Looks good now, thanks! I'll merge after the tests have gone green.

Issue #331, opened 2019-07-06
Bugs with `gaussian_blur_f32` when sigma <= 0.0


Aloxaf:
# Sigma == 0.0
`gaussian_blur_f32` will clean my image

```rust
use image::{RgbImage, Rgb};
use imageproc::filter::gaussian_blur_f32;
use imageproc::drawing::draw_filled_circle_mut;

fn main() {
    let mut image = RgbImage::new(200, 200);
    draw_filled_circle_mut(&mut image, (100, 100), 100, Rgb([255, 255, 255]));
    let _ = image.save("original.png");
    let _ = gaussian_blur_f32(&image, 0.0).save("zero.png");
}
```

original.png
![original](https://user-images.githubusercontent.com/17017672/60756609-f1146c80-a031-11e9-99ce-f5c0f60f0c9c.png)

zero.png
![zero](https://user-images.githubusercontent.com/17017672/60756612-f4a7f380-a031-11e9-9361-dfa6b8ee5073.png)

# Sigma <= - 1.0
`gaussian_blur_f32` will cause panic

## Backtrace

```
thread 'main' panicked at 'attempt to multiply with overflow', /home/aloxaf/.cargo/registry/src/github.com-1ecc6299db9ec823/imageproc-0.18.0/src/filter/mod.rs:157:37
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:212
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:475
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:382
   8: rust_begin_unwind
             at src/libstd/panicking.rs:309
   9: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  10: core::panicking::panic
[...]

PR #332, opened 2019-07-06
Enforce sigma > 0.0 in gaussian_blur_f32

theotherphil:
Fixes https://github.com/image-rs/imageproc/issues/331

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -166,12 +166,17 @@ fn gaussian_kernel_f32(sigma: f32) -> Vec<f32> {
 /// Blurs an image using a Gaussian of standard deviation sigma.
 /// The kernel used has type f32 and all intermediate calculations are performed
 /// at this type.
+///
+/// # Panics
+///
+/// Panics if `sigma <= 0.0`.
 // TODO: Integer type kernel, approximations via repeated box filter.
 pub fn gaussian_blur_f32<P>(image: &Image<P>, sigma: f32) -> Image<P>
 where
     P: Pixel + 'static,
     <P as Pixel>::Subpixel: ValueInto<f32> + Clamp<f32>,
 {
+    assert!(sigma > 0.0, "sigma must be > 0.0");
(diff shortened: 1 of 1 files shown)


src/filter/mod.rs
```diff
@@ -166,12 +166,17 @@ fn gaussian_kernel_f32(sigma: f32) -> Vec<f32> {
...
+/// # Panics
+///
+/// Panics if `sigma <= 0.0`.
 // TODO: Integer type kernel, approximations via repeated box filter.
 pub fn gaussian_blur_f32<P>(image: &Image<P>, sigma: f32) -> Image<P>
 where
     P: Pixel + 'static,
     <P as Pixel>::Subpixel: ValueInto<f32> + Clamp<f32>,
 {
+    assert!(sigma > 0.0, "sigma must be > 0.0");
```
upsuper: I suggest that `debug_assert` is enough, since this being violated wouldn't really lead to any memory safety issue, so it doesn't necessarily need a runtime enforcement on release build.

theotherphil: It wouldn't lead to memory unsafety, but it would lead to incorrect behaviour.

If panicking causes issues in practise then we could switch all of this library to using `Results` instead of panicking, but I'm a bit reluctant to do this as it makes all function calls a bit more verbose and I think that ensuring correctness from the caller should be straightforward.

upsuper: I agree that it is better just panic here, as the non-positive value just doesn't make sense as input here. But `assert` has runtime overhead (although not a lot) so ideally this should only be checked on debug build, where people can easily identify the issue when they hit it.

theotherphil: In this case the overhead of the check will be imperceptible, so I prefer to have consistent behaviour between release and debug builds.

HeroicKatora: Would also enforce it in both builds. I don't see any runtime impact in checking and consistent behaviour is easier to debug.

PR #345, opened 2019-07-26
Use canvas in draw_text

sinhpham:
Enable drawing semi transparent text.

### DIFF
--- a/src/drawing/canvas.rs
+++ b/src/drawing/canvas.rs
@@ -1,8 +1,4 @@
-use image::{
-    GenericImage,
-    GenericImageView,
-    Pixel
-};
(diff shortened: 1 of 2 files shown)


src/drawing/canvas.rs
```diff
@@ -68,22 +64,29 @@ pub trait Canvas {
         self.dimensions().1
     }
 
+    /// Returns the pixel located at (x, y)
```
theotherphil:
```suggestion
    /// Returns the pixel located at (x, y).
```

theotherphil:
One tiny nit but otherwise looks good, thanks!

A regression test would be nice, but I don’t we don’t have any for text drawing currently, so if you’ve tested manually that’ll do.

PR #350, opened 2019-08-13
Geometric Transformations Module

hepek:
Added a module named proj that defines projective transformations, and means to construct, combine and apply them to images.

The module exports warp* functions that allow performing projective transformations on images. Some of the functionality of this module overlaps with the affine module.

It exports a type Proj that can be used to define projective transformations by
 * Constructing them from primitive transformations
 * From a given 3x3 projective matrix (in homogeneous coordinates)
 * Or from a set of 4 control point pairs (for this calculation I have added a new dependency `rulinalg 0.4.2` - a linear algebra crate

image dependency bumped to 0.22.1 and all functions deprecated in the image crate marked with #[allow_deprecated] to suppress warnings.

Bumped the version of the imageproc crate. Don't know if this is common practice in additions like this one.


HeroicKatora:
Any reasoning behind the choice of the linalg crate? It seems to be fairly unmaintained and has not reached `1.0` yet.

theotherphil:
Thanks! I’ll review this properly on the weekend.

A quick note: I bumped the version of image in https://github.com/image-rs/imageproc/pull/347 and allowed deprecated functions across the whole crate. Please remove the per-function attributes you’ve added. Please also remove the imageproc version bump - I’ll do that as a separate PR and release a new version very soon.

hepek:
@HeroicKatora the only reasoning was that it does not depend on BLAS and LAPACK. I was looking into nalgebra, but that turned me down.

@theotherphil - deal. I will revert the per-function attributes later today.

HeroicKatora:
> @HeroicKatora the only reasoning was that it does not depend on BLAS and LAPACK.

That's actually perfectly fine reasoning. Its dependencies seem rather small for a llinalg library, it mainly depends on `num` crate. The `rustc-serialize` dependency is very dated but optional and not in default features. All is well.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+}
+
+/// A 2d Projective transformation, stored as a row major 3x3 matrix.
+/// Transformations combine by multiplication.
+///
+/// I.e. to define rotation around image center
+/// ```
+/// use imageproc::proj::*;
+/// let (cx,cy) = (320.0,240.0);
+/// let c_rotation = Proj::trans(cx,cy)*Proj::rot(30.0)*Proj::trans(-cx, -cy);
```
theotherphil: Nit: always insert spaces after commas in tuples, and around operators.

In this case `(320.0, 240.0)` and 'Proj::trans(cx, cy) * Proj::rot(30.0) * Proj::trans(-cx, -cy);`

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+    /// Matrix must be invertible, otherwise it does not define a Projection (by definition).
+    pub fn from_matrix(transform: [f32; 9]) -> Option<Proj> {
+        let transform = normalize(transform);
+        let class = class_from_matrix(transform);
+        try_inverse(&transform)
+            .map(|inverse| 
+                 Proj { transform, inverse, class })
+    }
+
+    /// Defines a translation by (tx, ty)
```
theotherphil:
```suggestion
    /// Defines a translation by (tx, ty).
```

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+            class: TransformationClass::Affine,
+        }
+    }
+
+    /// Inverts the transformation.
+    pub fn inv(self) -> Proj {
+        Proj { transform: self.inverse, inverse: self.transform, class: self.class }
+    }
+
+    /// Calculates a homography from a set of four control point pairs
```
theotherphil:
```suggestion
    /// Calculates a homography from a set of four control point pairs.
```

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+    pub fn from_cp(from: [(f32, f32); 4], to: [(f32, f32); 4]) -> Option<Proj> {
```
theotherphil: This abbreviation seems unnecessarily cryptic. `from_control_points`?

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+             from[3].0 as f64, from[3].1 as f64);
+
+        let (x1, y1, x2, y2, x3, y3, x4, y4) =
+            (to[0].0 as f64, to[0].1 as f64,
+             to[1].0 as f64, to[1].1 as f64,
+             to[2].0 as f64, to[2].1 as f64,
+             to[3].0 as f64, to[3].1 as f64);
+
+        let a = Matrix::new(9, 9, vec![
+            0f64,0.0, 0.0,-xf1,-yf1,-1.0, y1*xf1, y1*yf1, y1,
```
theotherphil:
```suggestion
            0f64, 0.0, 0.0, -xf1, -yf1, -1.0, y1 * xf1, y1 * yf1, y1,
```

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+                    let class = TransformationClass::Projection;
+
+                    try_inverse(&transform)
+                        .map(|inverse| { Proj { transform, inverse, class }})
+                }
+            }
+        }
+    }
+
+    // Helper functions used as optiomization in warp
```
theotherphil:
```suggestion
    // Helper functions used as optimization in warp
```

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+
+    #[inline(always)]
+    fn warp_p(&self, x: f32, y: f32) -> (f32, f32) {
```
theotherphil: Please give these functions clearer names.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+    }
+}
+
+/// Combines two projective transformations
+/// (matrix multiplication)
+impl<'a, 'b> Mul<&'b Proj> for &'a Proj {
+    type Output = Proj;
+
+    fn mul(self, rhs: &Proj) -> Proj {
+        use TransformationClass as TC;
```
theotherphil: You can define one of these two `Mul` implementations in terms of the other, and remove the duplication.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+/// Performs projective transformation to an image. Allocates an output image with same
+/// dimensions as the input image. Output pixels outside the input image are set to `default`.
+/// Proj `homography` defines a mapping from coordinates in the `output` image to coordinates of
+/// the `image`.
+pub fn warp_new<P>(homography: &Proj,
+                   image: &Image<P>,
+                   interpolation: Interpolation,
+                   default: P
+                  ) -> Image<P> 
+where
```
theotherphil: The formatting of the function parameters and the where clauses don't match the rate of this library.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+    let (width, height) = image.dimensions();
+    let mut out = ImageBuffer::new(width, height);
+    warp(homography, image, interpolation, default, &mut out);
+
+    out
+}
+
+/// Performs projective transformation `homography` mapping pixels from
+/// `image` into `&mut output`.
+/// Proj `homograpy` defines coordinate mapping from `out` to `image`.
```
theotherphil: This seems like the wrong way round...

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+pub fn warp<P>(homography: &Proj,
```
theotherphil: In the rest of this library we use `my_func` to name functions which allocate a new return value, and `my_func_mut` to refer to those that mutate an image in place. There aren't yet any any public functions that take a preallocated output, so we don't have a naming convention for this.

I suggest

`warp(...) -> Image`
`warp_into(..., &mut)`

but am open to better suggestions.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+/// use image::{ImageBuffer, Luma};
+/// use imageproc::utils::gray_bench_image;
+/// use imageproc::proj::*;
+///
+/// let img = gray_bench_image(300, 300);
+/// let mut out = ImageBuffer::new(300, 300);
+/// warp_with(|x,y| (x, y+(x/30.0).sin()), &img, Interpolation::Nearest, Luma([0u8]), &mut
+/// out);
+/// ```
+pub fn warp_with<P, Fm>(mapping: Fm,
```
theotherphil: Including this function in this module with this name is confusing - it has nothing to do with projections.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+
+    match interpolation {
+        I::Nearest => warp_inner(out, mapping, nn),
+        I::Bilinear => warp_inner(out, mapping, bl),
+    }
+}
+
+fn warp_inner<P,Fc,Fi>(
+    out: &mut Image<P>,
+    coord_tf: Fc,
```
theotherphil: This parameter is called `mapping` in the previous function, which is more understandable.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+}
+
+fn normalize(mx: [f32; 9]) -> [f32; 9] {
+    [mx[0]/mx[8], mx[1]/mx[8], mx[2]/mx[8],
+     mx[3]/mx[8], mx[4]/mx[8], mx[5]/mx[8],
+     mx[6]/mx[8], mx[7]/mx[8], mx[8]/mx[8]]
+}
+
+fn try_inverse(t: &[f32; 9]) -> Option<[f32; 9]> {
+    let (
```
theotherphil: This is copy-pasted from the equivalent function from `affine`, with some minor tweaks. Please extract most of the logic into a common function that they can share.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+        b[3], b[4], b[5], 
+        b[6], b[7], b[8]
+        );
+
+    [a11*b11 + a12*b21 + a13*b31, a11*b12 + a12*b22 + a13*b32, a11*b13 + a12*b23 +a13*b33,
+     a21*b11 + a22*b21 + a23*b31, a21*b12 + a22*b22 + a23*b32, a21*b13 + a22*b23 + a23*b33,
+     a31*b11 + a32*b21 + a33*b31, a31*b12 + a32*b22 + a33*b32, a31*b13 + a32*b23 + a33*b33]
+}
+
+fn blend<P>(
```
theotherphil: This is copy-pasted without modification from `affine`, right? Reuse that function here.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+    });
+
+    top.map2(&bottom, |u, v| {
+        P::Subpixel::clamp((1f32 - bottom_weight) * cast(u) + bottom_weight * cast(v))
+    })
+}
+
+
+#[allow(deprecated)]
+fn interpolate<P>(image: &Image<P>, x: f32, y: f32, default: P) -> P
```
theotherphil: And this one?

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+                image.unsafe_get_pixel(right as u32, bottom as u32),
+            )
+        };
+        blend(tl, tr, bl, br, right_weight, bottom_weight)
+    }
+}
+
+#[inline(always)]
+#[allow(deprecated)]
+fn nearest<P: Pixel + 'static>(image: &Image<P>, x: f32, y: f32, default: P) -> P {
```
theotherphil: This one is slightly different to the equivalent function in `affine`, but I'm not sure why.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+    } else if rx >= width || ry >= height {
+        default
+    } else {
+        unsafe { image.unsafe_get_pixel(rx, ry) }
+    }
+}
+
+/// How to handle pixels whose pre-image lies between input pixels.
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum Interpolation {
```
theotherphil: Use the type from `affine`.

Maybe combine affine.rs and proj.rs into a new geometric_transformation.rs? We should keep the specialised functions for translation, rotation, etc., as we can make them faster than a general projective transformation, but I'm not sure that we then need a special case for Affine transformations - those can be subsumed into your new projection functions.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+    Nearest,
+    /// Bilinearly interpolate between the four pixels
+    /// closest to the pre-image of the output pixel.
+    Bilinear,
+}
+
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
theotherphil: We should remove duplication between these tests and those in `affine`. If you merge the two files then we can at least put the equivalent tests for `some_specialised_function` and `equivalent_projective_transformation` next to each other.

src/proj.rs
hepek: I agree. Will do in my next update.

src/proj.rs
hepek: Yup. Didn't want to make it public in the affine module, and I was hoping to replace it with something faster, but couldn't achieve better perf with a couple of simd crates I have tried.

src/proj.rs
hepek: The change was made during performance optimizations, but, to be honest, I am not sure it actually is any faster with this change any more.

src/proj.rs
hepek: I like this suggestion. Thought about that before, but didn't want to seem too pushy, hence the separate module. I will combine the two in geometric_trasnformation.rs.

That will make a breaking change in the crate, so I guess after that a version bump is really needed.

theotherphil: Thanks. Yes, I’ll publish a new version after this PR is landed.

src/proj.rs
hepek: @theotherphil, this is common in many image processing libraries (check out opencv's docs for example: https://docs.opencv.org/2.4/modules/imgproc/doc/geometric_transformations.html?highlight=warp).

Since getting the inverse transformation from Proj is super cheap, it could be either way. But if we want to keep functions warp_with that allow passing a custom function, than that could be rather confusing to have two different conventions in the same crate. 

What do you think?

src/proj.rs
hepek: I like warp and warp_into.

src/proj.rs
hepek: Done.

src/proj.rs
HeroicKatora: OpenCV isn't really the cleanest reference to be honest, I always found it super weird and piecing together transformations without the reference close to impossible. Though it is prior art...

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+/// Transformations combine by multiplication.
+///
+/// I.e. to define rotation around image center
+/// ```
+/// use imageproc::proj::*;
+/// let (cx,cy) = (320.0,240.0);
+/// let c_rotation = Proj::trans(cx,cy)*Proj::rot(30.0)*Proj::trans(-cx, -cy);
+/// ```
+#[derive(Copy, Clone, Debug)]
+pub struct Proj {
```
HeroicKatora: Why not name this `Projection`?

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+    pub fn from_matrix(transform: [f32; 9]) -> Option<Proj> {
+        let transform = normalize(transform);
+        let class = class_from_matrix(transform);
+        try_inverse(&transform)
+            .map(|inverse| 
+                 Proj { transform, inverse, class })
+    }
+
+    /// Defines a translation by (tx, ty)
+    pub fn trans(tx: f32, ty: f32) -> Proj {
```
HeroicKatora: Rename to `translate`.

src/proj.rs
```diff
@@ -0,0 +1,791 @@
...
+                0.0, 1.0, -ty,
+                0.0, 0.0, 1.0],
+            class: TransformationClass::Translation,
+        }
+    }
+
+    /// Defines a rotation around the origin by angle theta degrees. Origin is (0,0) pixel
+    /// coordinate, usually top left corner. To rotate around image center combine rotation with 
+    /// two translations: T*Rotation*T^-1.
+    pub fn rot(theta: f32) -> Proj {
```
HeroicKatora: Rename to rotate and document the direction of rotation. Since you already give a formula for rotation around center, maybe you should add a utility method?

src/proj.rs
HeroicKatora: Maybe this can be resolve with naming:

* `warp`: takes only `Proj` but 'the right way around'. Parameter order: `(image, Proj, Interpolation, out)`
* `mapped_from`: the arbitrary mapping and thus reverse projection. other parameter order
   `(out, projection, Interpolation, image)`

Just an idea

src/proj.rs
hepek: Renamed to map_projective, map_affine, map_translation

src/proj.rs
hepek: translate or translation?

HeroicKatora: It should be a verb, `translate`.

src/proj.rs
hepek: Not bad. I was playing with the idea, but reverted to the `accepted` order after I added warp_with.

src/proj.rs
hepek: Will be better when the module name becomes geometric_transformations.

src/proj.rs
hepek: Fixed. That was sloppy, sorry.

HeroicKatora:
Idea for a clearer and cleaner interface for warping. A wrapper for an image that embeds them into a coordinate system. The mapping would describe a viewport in that single coordinate system. It would also solve the issue of having find the center of an image only in documentation and being burdened with creating all matrices.

I know this would not capture the arbitrary function use case but I don't think it should be pulicly visiable in a module called for 'projective transformations' anyways.

Sketch:

```
struct Projected<P> {
    image: Image<P>,
    /// Describes mapping from the trivial independent set of the homography 
    /// to the image corners. But this is an implementation detail, could also be
    /// the inverse matrix.
    viewport: Projection,
}

impl<P> Projected<P> {
    /// Embed an image at its pixel coordinates.
    /// Coordinates (x, y) correspond to pixels (x, y).
    pub fn new(image: I) -> Self { … }
    /// Embed an image into the unit coordinates.
    /// The top-left pixel is mapped to (0, 0), the top-right corner is at (0, 1),
    /// the bottom left at (1, 0), and the bottom right at (1, 1).
    pub fn into_unit_square(image: I) -> Self { … }
    // … constructors with provided Projection.

    // … scale, translate, rotate

    /// Assigns image pixel from a reference if they are in-bounds.
    pub fn overlay(&mut self, from: &Projected<P>);
}
```

src/proj.rs
hepek: @theotherphil, finished merging affine and proj module. Took affine::translate verbatim, as it was super fast. Reimplemented rotate by using Projection + warp.
Removed most of the helper functions that play around HasBlack trait.

I wanted to use try_invert with f64 - for better numeric stability, but it was crashing regression tests (affine_bilinear_rgb) so I gave up on that.

- warp* with projection take in->out homography as was the case with affine.
- reverted order of parameters so in image is first, and homography is 2nd (to be more like affine)

theotherphil:
Thanks for the changes. I'm afraid I have very little time during the week, but I hope to get round to this before the weekend this time!

theotherphil:
I've left comments for a couple of small issues. There's quite a lot of tidying of formatting and documentation that I'd like to do, but instead of you making changes and then waiting forever for me to do another review I'll merge this PR after you've re-added `rotate_about_center` and removed the `println`s, and then I'll tinker with this file a bit before publishing a new crate version.

hepek:
Ok. Sounds like a plan. I'll make those changes in a couple of hours.

hepek:
Ok. Done.

theotherphil:
Thanks!

theotherphil:
@hepek Your changes are in imageproc 0.19, along with a bump to the referenced version of the image crate.

hepek:
Thank you.

ned, 25. avg 2019. 13:30 theotherphil <notifications@github.com> je
napisao/la:

leo60228:
> the only reasoning was that it does not depend on BLAS and LAPACK. I was looking into nalgebra, but that turned me down.

`nalgebra-lapack` is a separate crate from `nalgebra`. `nalgebra` is pure Rust, and it looks like it's suitable for this use-case. I'm working on a PR.

hepek:
@leo60228, there is also a bunch of geometry primitives in nalgebra
https://nalgebra.org/rustdoc/nalgebra/geometry/index.html similar to what is used here.
I hope you'll include that in your PR too.

leo60228:
I was trying to avoid a breaking change, and those types have a very different API from imageproc.

leo60228:
#397

Issue #343, opened 2019-07-18
Use cargo fmt?


windelbouwman:
I just formatted my local copy with the command `cargo fmt`, which formats all sourcecode. It then appeared to me, that I changed a lot of files :).

Is it the goal to format all files with cargo fmt? One issue I see with this, is that matrices are not nicely formatted anymore. But then again, this rustfmt tool is really handy.

theotherphil:
The files in this library were all run through rustfmt at some point, but the defaults have changed since. I'd happily accept PRs to update the formatting, but I'd like to maintain the formatting of the matrices.

JohnTitor:
FWIW: this config is now outdated.

https://github.com/image-rs/imageproc/blob/4cf5abb1f82f0e8e929c35e66df80b42617dc643/rustfmt.toml#L1

PR #359, opened 2019-08-25
Format code by rustfmt

JohnTitor:
This is a part of #343.
I'll submit PRs little by little because the diff will be too large to review on GitHub.
And at first, we should consider formatting rules as @theotherphil mentioned in https://github.com/image-rs/imageproc/issues/343#issuecomment-512971336.


src/drawing/rect.rs
```diff
@@ -124,20 +124,22 @@ mod tests {
...
         let mut blended = blue;
         blended.blend(&semi_transparent_red);
 
         let expected = vec![
-            white, white,  white,  white, white,
-            white,  blue,   blue,   blue, white,
-            white,  blue, blended,  blue, white,
-            white,  blue,   blue,   blue, white,
-            white, white,  white,  white, white
+            white, white, white, white, white, white, blue, blue, blue, white, white, blue,
```
theotherphil: This should remain formatted as a 5x5 grid.

src/filter/median.rs
```diff
@@ -242,12 +256,15 @@ impl HistSet {
             data.push([0u32; 256]);
         }
 
-        HistSet { data, expected_count }
+        HistSet {
```
theotherphil: What's the rule that's being applied here? This line is already very short. Does rustfmt now enforce one row per field for all struct initialisations?

src/filter/mod.rs
```diff
@@ -692,10 +692,7 @@ mod tests {
 
     #[test]
     fn test_filter3x3_with_results_outside_input_channel_range() {
-        let kernel: Vec<i32> = vec![
-            -1, 0, 1,
-            -2, 0, 2,
-            -1, 0, 1];
+        let kernel: Vec<i32> = vec![-1, 0, 1, -2, 0, 2, -1, 0, 1];
```
theotherphil: This should remain a 3x3 grid.

src/filter/mod.rs
```diff
@@ -754,16 +751,10 @@ mod tests {
             9, 4;
             8, 1);
 
-        let k: Vec<f32> = vec![
-            0.1, 0.2, 0.1,
-            0.2, 0.4, 0.2,
-            0.1, 0.2, 0.1
-        ];
+        let k: Vec<f32> = vec![0.1, 0.2, 0.1, 0.2, 0.4, 0.2, 0.1, 0.2, 0.1];
```
theotherphil: And this

src/filter/mod.rs
```diff
@@ -775,10 +766,7 @@ mod tests {
     #[bench]
     fn bench_filter3x3_i32_filter(b: &mut Bencher) {
         let image = gray_bench_image(500, 500);
-        let kernel: Vec<i32> = vec![
-            -1, 0, 1,
-            -2, 0, 2,
-            -1, 0, 1];
+        let kernel: Vec<i32> = vec![-1, 0, 1, -2, 0, 2, -1, 0, 1];
```
theotherphil: And this

tests/regression.rs
```diff
@@ -136,52 +166,98 @@ fn test_equalize_histogram_grayscale() {
...
 fn test_affine_nearest_rgb() {
     fn affine_nearest(image: &RgbImage) -> RgbImage {
-        let root_two_inv = 1f32/2f32.sqrt()*2.0;
+        let root_two_inv = 1f32 / 2f32.sqrt() * 2.0;
         let hom = Projection::from_matrix([
-            root_two_inv, -root_two_inv,  50.0,
-            root_two_inv,  root_two_inv, -70.0,
-            0.0         , 0.0          , 1.0,
-        ]).unwrap();
+            root_two_inv,
```
theotherphil: And this

tests/regression.rs
```diff
@@ -136,52 +166,98 @@ fn test_equalize_histogram_grayscale() {
...
 fn test_affine_bilinear_rgb() {
     fn affine_bilinear(image: &RgbImage) -> RgbImage {
-        let root_two_inv = 1f32/2f32.sqrt()*2.0;
+        let root_two_inv = 1f32 / 2f32.sqrt() * 2.0;
         let hom = Projection::from_matrix([
-            root_two_inv, -root_two_inv,  50.0,
-            root_two_inv,  root_two_inv, -70.0,
-            0.0         , 0.0          , 1.0,
-        ]).unwrap();
+            root_two_inv,
```
theotherphil: And this

src/filter/median.rs
JohnTitor: Uh, indeed, it's aggressive a bit. I think the reason is here https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#struct. I added `#[rustfmt::skip]`

JohnTitor:
I added `#[rustfmt::skip]` for grid styles.

src/filter/median.rs
```diff
@@ -242,12 +256,13 @@ impl HistSet {
             data.push([0u32; 256]);
         }
 
+        #[rustfmt::skip]
         HistSet { data, expected_count }
```
theotherphil: Sorry to be awkward, but I don't think it makes sense to opt out of the standard formatting for this one line - I was just a bit surprised that rustfmt decided to split this line.

Please could you revert this line and then I'll merge the PR.

JohnTitor: Okay, I reverted the change.

src/filter/median.rs
theotherphil: Thanks for the link. I think we should just go with the default here, even if it's a bit odd.

theotherphil:
Thanks!

PR #363, opened 2019-08-26
Format code by rustfmt (part 4)

JohnTitor:
This is a final step for #343.

One question: we should run rustfmt on CI? If so, I'd happy to submit a PR to do it. If no, all we have to do is close the issue.


theotherphil:
Thanks.

I'd be happy for incorrect formatting to cause the tests to fail (as done in e.g. rust-analyzer), but I'd like to avoid automatically reformatting code after PRs are merged.

PR #365, opened 2019-08-26
Run rustfmt on stable on Travis

### LINKED ISSUE #343: Use cargo fmt?

JohnTitor:
Fixes #343 
I think we don't have to run rustfmt for all targets so make it run only stable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,19 @@
 language: rust
 
 rust:
-    - stable
-    - beta
-    - nightly
+  - stable
+  - beta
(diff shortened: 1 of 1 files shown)


theotherphil:
Thanks!

PR #375, opened 2019-09-01
Add comment justifying unsafety in corner.rs, fix bug for very large …

theotherphil:
…or very small images


HeroicKatora:
Could this lead to actual out-of-bounds reads?

PR #378, opened 2019-09-01
Fix to point to image-rs

JohnTitor:


JohnTitor:
One minor question: is this still necessary?
https://github.com/image-rs/imageproc/blob/master/scripts/travis-doc-upload.cfg

theotherphil:
Thanks! And no, that is no longer necessary.

Issue #358, opened 2019-08-25
Docs failed to build for 0.19


theotherphil:
https://docs.rs/crate/imageproc/0.19.0/builds/180949

theotherphil:
And 0.19.1. I've raised https://github.com/rust-lang/docs.rs/issues/408 to see if there's an easy workaround.

theotherphil:
I've disabled the display-window feature and published 0.19.2. We now have docs for the rest of the library.

ripytide:
Looks like this issue can be closed now too.

theotherphil:
There's still the issue of missing docs for code behind the display-window feature.

ripytide:
Is there any way of solving that issue? `sdl2` requiring internet access and `docs.rs` not allowing it seem quite opposed.

PR #379, opened 2019-09-01
Remove unnecessary file

JohnTitor:


theotherphil:
Ah, it's still used in https://github.com/image-rs/imageproc/blob/master/.travis.yml. It can be deleted there too, as we're not pointing to the docs on docs.rs (although they're not currently building... https://github.com/image-rs/imageproc/issues/358)

JohnTitor:
Also removed from Travis

theotherphil:
Thanks!

PR #386, opened 2019-09-13
fix #383

### LINKED ISSUE #383: connected_components on wasm fails at size check
### Bug

In current HEAD, and also in `0.18.0` (which is the version I am using), `imageproc::region_labelling::connected_components` does the following check:

```rust
if image_size >= 2usize.pow(32) {
  panic!("Images with 2^32 or more pixels are not supported");
}
```

However `usize` on wasm is 32 bits which means that `2usize.pow(32)` overflows and becomes `0`, so the check ends up being `if image_size >= 0` so fails on all images.

### Possible fix

If I understand the intent of this check correctly, it is a protection against usage of `image_size` later as the size of an array:

```rust
let mut output_labels = vec![0u32; image_size];
```

I think we can get the same protection by changing the check to:

```rust
if image_size >= usize::max_value() as u32 {
  panic!(
    "Images with more pixels than `usize` supports ({}) are not supported",
    usize::max_value()
  );
}
```

I've started on this approach here: https://github.com/mikemoraned/imageproc/pull/1 . If this fix/approach is acceptable then I'll turn it into a full PR against this repo and submit as a fix for this bug.

mikemoraned:
fix #383 by:
- extending tests for `region_labelling` to run on `wasm32`
- changing implementation of size check in `connected_components` to work on wasm32

### DIFF
--- a/src/region_labelling.rs
+++ b/src/region_labelling.rs
@@ -244,13 +244,18 @@ where
 
 #[cfg(test)]
 mod tests {
+    extern crate wasm_bindgen_test;
+
     use super::connected_components;
     use super::Connectivity::{Eight, Four};
     use crate::definitions::{HasBlack, HasWhite};
     use ::test;
     use image::{GrayImage, ImageBuffer, Luma};
+    #[cfg(target_arch = "wasm32")]
+    use wasm_bindgen_test::*;
 
-    #[test]
+    #[cfg_attr(not(target_arch = "wasm32"), test)]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn test_connected_components_eight_white_background() {
         let image = gray_image!(
               1, 255,   2,   1;
@@ -280,7 +285,8 @@ mod tests {
         })
     }
 
-    #[test]
+    #[cfg_attr(not(target_arch = "wasm32"), test)]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn test_connected_components_eight_chessboard() {
         let image = chessboard(30, 30);
         let components = connected_components(&image, Eight, Luma::black());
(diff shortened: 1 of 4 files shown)


src/region_labelling.rs
```diff
@@ -128,8 +128,8 @@ where
     I::Pixel: Eq,
 {
     let (width, height) = image.dimensions();
-    let image_size = width as usize * height as usize;
-    if image_size >= 2usize.pow(32) {
+    let image_size: usize = width as usize * height as usize;
```
theotherphil: Why have you added the type annotation here?

src/region_labelling.rs
```diff
@@ -244,13 +244,18 @@ where
...
     use crate::definitions::{HasBlack, HasWhite};
     use ::test;
     use image::{GrayImage, ImageBuffer, Luma};
+    #[cfg(target_arch = "wasm32")]
+    use wasm_bindgen_test::*;
 
-    #[test]
+    #[cfg_attr(not(target_arch = "wasm32"), test)]
+    #[cfg_attr(target_arch = "wasm32", wasm_bindgen_test)]
     fn test_connected_components_eight_white_background() {
```
theotherphil: This is a lot of work to annotate three tests. Does wasm-bindgen-test require annotating every single test you want to run this way, and are you expecting to increase the set of tests that gets run this way over time?

src/region_labelling.rs
mikemoraned: It was mostly be making sure the types were correct; I've removed it now as it's unnecessary.

src/region_labelling.rs
mikemoraned: Yes, agreed; this is quite verbose. And yes, wasm-bindgen-test needs this. However, ideally it wouldn't, so I'm going to do a bit more research to see if it is definitely needed and/or if there is something less verbose.

Re: "expecting to increase the set of tests that gets run this way over time", it would make sense to, so that we have full wasm coverage. Separate to this, I've started working on full wasm coverage, see: https://github.com/mikemoraned/imageproc/pull/3 . This has shown up another couple of bugs, I think related to OsRng usage. However, I'll open a separate PR for anything that comes from this.

theotherphil: Ok, thanks for the information. I'm fine to merge this as-is.

PR #389, opened 2019-09-30
Update dependencies

JohnTitor:
Update `itertools` and `rusttype`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ script:
   - |
     if [ "$TRAVIS_RUST_VERSION" = "nightly" ]; then
       rustup target add wasm32-unknown-unknown --toolchain nightly
-      cargo install wasm-bindgen-cli --vers "0.2.50"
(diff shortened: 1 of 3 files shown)


theotherphil:
The build failure is in the recently added wasm build.

error: executing `wasm-bindgen` over the wasm file
	caused by: 
it looks like the Rust project used to create this wasm file was linked against
a different version of wasm-bindgen than this binary:
  rust wasm file: 0.2.51
     this binary: 0.2.50
Currently the bindgen format is unstable enough that these two version must
exactly match, so it's required that these two version are kept in sync by
either updating the wasm-bindgen dependency or this binary. You should be able
to update the wasm-bindgen dependency with:
    cargo update -p wasm-bindgen
or you can update the binary with
    cargo install -f wasm-bindgen-cli
if this warning fails to go away though and you're not sure what to do feel free
to open an issue at https://github.com/rustwasm/wasm-bindgen/issues!

JohnTitor:
Resolved

theotherphil:
Thanks!

PR #390, opened 2019-10-19
Fix translate() which panics on negative x offset

### LINKED ISSUE #384: translate bug
```
let image = image::RgbImage::new(100, 200);
let translated = translate(&image, (-10, -20));
```

`thread 'geometric_transformations::tests::test_translate_sadness' panicked at 'attempt to subtract with overflow', src/geometric_transformations.rs:357:31`

sjaustirni:
Fixes #384 

Adds a unit test and a regression test as well.

If regression test is not desired, I can just revert that commit.


theotherphil:
Thanks. Could you please remove the regression test, as I don't think it provides much in this case.

sjaustirni:
@theotherphil I was contemplating about the same thing. I included it just in case anyway, since I already had it (to visually confirm that the change works).

The commit has been reverted now.

theotherphil:
Thanks!

PR #391, opened 2019-10-20
WIP: Implement rotate_uncropped()

### LINKED ISSUE #323: rotate without cropping
Currently, rotating crops the parts of an image which are "rotated out" of the image buffer (the corners). While this makes sense (resizing the buffer would be necessary), a function which does the resizing and rotation in one step would be super usefull.

Right now, a naive solution to achive this could look something like [this](https://pastebin.com/PJznTwFK), but it's *really* slow, for obvious reasons.

sjaustirni:
Fixes #323 

Implements `rotate_uncropped()`, which rotates the original image passed in and returns a new image, which contains the whole rotated original image centered in this new image - in other words; the function rotates the image and returns it uncropped.

Thank you @NyxCode for the original inspiration.

As always, if the regression test is undesired, it can be removed :)


theotherphil:
Thanks for the PR. This is definitely a useful function to add.

The image you're committing for the regression test appears to have a black bar running down the left hand side. Do you know why this is?

Could you please also add a regression test for a more interesting angle of rotation.

sjaustirni:
@theotherphil Good catch! My image viewer has blackish background, so I did not see it initially.

It seems that the black bars are more prevalent, there is also a horizontal one if you rotate by correct amount (see b84c1b587fad1df109e833dc53d455d4c84eab16).

I'll try to resolve it, but currently I do not know what the underlying issue is.

PR #394, opened 2019-11-22
added l1 norm in math

dimitriGallos:
Simply added the l1 norm calculation


theotherphil:
Thanks! (And sorry for the long delay.)

PR #395, opened 2019-11-25
WIP: add sliding window iterator 

flappyBug:
Sliding window iterator over an image can have broad usages in image process. For example, we can manage filter easily using sliding window iterator without worrying about loop index. Getting gradient of a image won't have to create two temporary image gx and gy with the aiding of sliding window. 

Currently we have a `window` module, for avoiding of ambiguity, I renamed it to `gui`. Any better naming suggestions are welcome.

The `SizeOption` naming conventions are the same with [Octave's conv2: shape](https://octave.sourceforge.io/octave/function/conv2.html), and the `PadOption`'s description can be find in [imfilter](https://octave.sourceforge.io/image/function/imfilter.html)

This is a work in progress implementation, and I am a rust newbie, it's grateful to have review from you.
- [x] Basic implementation
- [ ] Document
- [ ] Test
- [ ] Use sliding window to refactor some of existing filter methods
- [ ] Benchmark against previous approach.

### DIFF
--- a/src/gui.rs
+++ b/src/gui.rs
@@ -1,0 +1,242 @@
+//! Displays an image in a window created by sdl2.
+
+use image::{imageops::resize, RgbaImage};
+use sdl2::{
+    event::{Event, WindowEvent},
+    keyboard::Keycode,
+    pixels::{Color, PixelFormatEnum},
+    rect::Rect,
+    render::{Canvas, TextureCreator, WindowCanvas},
+    surface::Surface,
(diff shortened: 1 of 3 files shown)


theotherphil:
Sorry for the long delay in replying to this PR.

The idea here seems useful, and like something that would make some functions easier to write. However, I'd like to see some numbers for the performance overhead of Window::get_pixel before reimplementing any existing functions using `Window`s.

PR #397, opened 2019-12-12
Switch from unmaintained rulinalg to nalgebra

leo60228:
`rulinalg` was initially used due to the belief that the `nalgebra-lapack` (which requires the system to have LAPACK) crate would need to be used instead of the pure-Rust `nalgebra`. However, this is incorrect, as nalgebra 0.13.0 added a pure-Rust SVD.


theotherphil:
Sorry for the long delay in responding to this PR.

This change increases debug build time from 28s to 49s on my machine. As we only use a single function from rulinalg I'm not concerned about it no longer being actively maintained.

I'll close without merging unless there's an argument that I'm missing in favour of this change.

PR #398, opened 2019-12-12
Fix class optimization

### LINKED ISSUE #393: Different behaviour from warp and warp_with
# Versions
image 0.22.3
imageproc 0.19.2
rustc 1.40.0-nightly (1423bec54 2019-11-05)

# Description
I'm encountering some really strange behaviour from `geometric_transformations::warp`.

For example, take this input image.
input

It contains a rectangle that's been warped by a perspective transform, and I'd like to invert this transform to obtain the original rectangle. I know the (rough) position of the rectangle's corners, so I use `Projection::from_control_points` and `warp`:
```rust
use image::Rgb;
use imageproc::geometric_transformations::{warp, Interpolation, Projection};
type Image = imageproc::definitions::Image<Rgb<u8>>;

fn extract_rect(image: Image) -> Image {
    let w = image.width() as f32;
    let h = image.height() as f32;

    let src = [(17.0, 20.0), (640.0, 20.0), (571.0, 382.0), (75.0, 382.0)];
    let dst = [(0.0, 0.0), (w, 0.0), (w, h), (0.0, h)];

    let p = Projection::from_control_points(src, dst)
        .expect("Could not compute projection matrix")
        .invert();

    warp(
        &image,
        &p,
        Interpolation::Bilinear,
        Rgb([0, 0, 0]),
    ))
}
```
However, the output image is not what I expect, as it just looks like the input image but slightly translated:

warp

What's more interesting, I've managed to obtain the result I was looking for by replacing `warp` with `warp_with`, and explicitly multiplying points with the transform:
```rust
use image::Rgb;
use imageproc::geometric_transformations::{warp_with, Interpolation, Projection};
type Image = imageproc::definitions::Image<Rgb<u8>>;

fn extract_rect(image: Image) -> Image {
    let w = image.width() as f32;
    let h = image.height() as f32;

    let src = [(17.0, 20.0), (640.0, 20.0), (571.0, 382.0), (75.0, 382.0)];
    let dst = [(0.0, 0.0), (w, 0.0), (w, h), (0.0, h)];

    let p = Projection::from_control_points(src, dst)
        .expect("Could not compute projection matrix")
        .invert();

    warp_with(
        &image,
        |x, y| p * (x, y),
        Interpolation::Bilinear,
        Rgb([0, 0, 0]),
    ))
}
```
Now the output image is this:

warp_with

Is `warp` applying the transform incorrectly, or am I misunderstanding how to use it?

leo60228:
Previously, it was only applied in some scenarios, and didn't work even then. Fixes #393.

### DIFF
--- a/src/geometric_transformations.rs
+++ b/src/geometric_transformations.rs
@@ -260,7 +260,11 @@ impl<'a, 'b> Mul<&'b (f32, f32)> for &'a Projection {
 
     fn mul(self, rhs: &(f32, f32)) -> (f32, f32) {
         let (x, y) = *rhs;
-        self.map_projective(x, y)
+        match self.class {
+            TransformationClass::Translation => self.map_translation(x, y),
+            TransformationClass::Affine => self.map_affine(x, y),
+            TransformationClass::Projection => self.map_projective(x, y),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


theotherphil:
Thanks! (And sorry for the delay.)

PR #405, opened 2020-01-19
Random color palette utils for visual debugging/visualization purposes

colbyn:
Alright results are looking interesting I think
Dy-b6Y3
Dy-b6Y3 debug

Using this:
```rust
pub fn preprocess(image: &DynamicImage) -> DynamicImage {
    // let image = image.resize_exact(700, 700, ::image::FilterType::Lanczos3);
    let image = image.to_luma();
    let image = imageproc::contrast::equalize_histogram(&image);
    let image = imageproc::filter::sharpen_gaussian(&image, 12.0, 38.0);
    let image = imageproc::region_labelling::connected_components(
        &image,
        Connectivity::Eight,
        Luma::black()
    );
    let image = color_palette::filter_luma_u32_regions(&image, 100);
    let image = image.to_pretty_rgb_palette();
    DynamicImage::ImageRgb8(image)
}
```

### DIFF
--- a/src/debug/color_palette.rs
+++ b/src/debug/color_palette.rs
@@ -1,0 +1,191 @@
+//! Color palette related visualization utilities.
+mod colourado;
+
+use std::collections::{HashMap};
+use image::{ImageBuffer};
+use image::{Luma, Rgb, Pixel};
+
+
+#[doc(hidden)]
+macro_rules! random_color_map {
+    ($image:expr) => {{
+        use colourado::{ColorPalette, PaletteType};
+        let keys: Vec<_> = $image
+            .pixels()
+            .map(|p| p[0]).collect();
+        let palette = ColorPalette::new(keys.len() as u32, PaletteType::Random, false);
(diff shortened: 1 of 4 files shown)


colbyn:
Not sure how useful this would be apart from visualizing the results from `connected_components`. Overall maybe the colors could perhaps be a bit more diverse...

CK-tgYJ debug
eH-vWUx debug

theotherphil:
Thanks for the PR. I don't have time at the moment to look into it,  but I'll get to it in the next couple of weeks. Sorry for the delay.

colbyn:
@theotherphil oh okay no rush.

ripytide:
I think this functionality is not a core image processing operation and so does not belong in the `imageproc` crate. I think the best place for such functionality would be in a separate crate called something like `image_debug_utils`, I'm not sure if such a crate already exists though. If the other crate is created then we could definitely link to it in the readme as well.

As such I would suggest closing this PR.

theotherphil:
Debugging utils are fine for this crate, but I agree that this functionality is a bit too niche.

PR #411, opened 2020-03-07
Make display_image() and display_multiple_images() generic

### LINKED ISSUE #408: The image argument of display_image is unnecessarily strict
As mentioned in title, the [display_image](https://github.com/image-rs/imageproc/blob/da34fbd27fc2a9d0c6daa1c9aff06d801242f32a/src/window.rs#L18) accepts `RgbaImage` rather than image generics. It restricts the acceptable image types.

Before the fix arrives, can anyone tell how to create owned image from image atop slices?

jerry73204:
This PR fixes #408 that `display_image()` only accepts `RgbaImage`. Now the argument is changed to generic that implements `GenericImageView`. Also, it fixes the `FilterType` not found error due to image crate update to 0.23.

### DIFF
--- a/src/window.rs
+++ b/src/window.rs
@@ -1,6 +1,9 @@
 //! Displays an image in a window created by sdl2.
 
-use image::{imageops::resize, RgbaImage};
+use image::{
+    imageops::{resize, FilterType},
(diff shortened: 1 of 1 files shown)


theotherphil:
Thanks!

PR #413, opened 2020-05-04
Fix display-window feature due to upstream refactoring of buffers

nelsonjchen:
Unfortunately the CI system doesn't really check this which is why this kind of went bad.


theotherphil:
Thanks. Please can you run 'cargo fmt' to resolve the build failure.

nelsonjchen:
@theotherphil should be done. 🤞 on check.

nelsonjchen:
Ick, screwed up my `git`. Hold on.

PR #414, opened 2020-05-04
Fix display-window feature due to upstream refactoring of buffers

nelsonjchen:
Sorry for the noise in #413. I am just having a really bad night with `git`. I would reopen it if I could.


theotherphil:
I know the feeling! Thanks for the fix.

PR #416, opened 2020-05-20
Update rusttype to latest version.

cellfusion:
An error sometimes occurred while reading the otf file, so the rusttype was updated to the latest version.


theotherphil:
Thanks!

Issue #312, opened 2019-02-01
Refactoring Template Matching Discussion


JonBoyleCoding:
When looking at the template matching to add cross-correlation (#311) I came up with a few things that might be an idea for refactoring; however, I thought it would be worth making a discussion before actually doing the work if it's not going to be wanted.

- Each of the template matching in https://docs.opencv.org/3.4/de/da9/tutorial_template_matching.html has exactly the same normalization term.
  - Would it be better (wanted) to have a separate argument to the template match function to specify whether to normalize? Otherwise the `MatchTemplateMethod` is going to have a lot of duplications and the various checks for methods are going to be increased two-fold whereas it could just become one function parameter.

- Currently the method can only accept single-channel grayscale images. The conversion to grayscale isn't always the same and isn't necessary for template matching in general. I've messed around with converting the function to accept any number of channels and data types and it should be possible, although will definitely complicate the code.

theotherphil:
Both of your suggestions sound good to me.

JonBoyleCoding:
I finally found some time to work on this. I have a working version of the second point, taking a GenericImageView as an argument instead of grayscale (works with current tests, need more for different types), however there is an issue with using the ArrayData::DataType that is returned from sum_image_pixels as the compiler can't seem to index into it when Pixel type is a generic.

I made a workaround for the moment; however, it changes the API of sum_image_pixels in integral_image to turn the ArrayData::DataType back into the pixel it is from: https://github.com/JonBoyleCoding/imageproc/tree/template_matching_update

I was wondering if you had any thoughts on how this might be best approached.

JonBoyleCoding:
I thought I would link to the actual files.

https://github.com/JonBoyleCoding/imageproc/blob/template_matching_update/src/template_matching.rs
https://github.com/JonBoyleCoding/imageproc/blob/template_matching_update/src/integral_image.rs

If you would rather I did this via pull request, I can do that as well.

PR #423, opened 2020-07-20
Modified Template Matching / Integral Image to use Image instead of GrayImage

JonBoyleCoding:
I decided to add a pull request to make it easier to see the changes discussed in #312. 

I had to change some API to make this work, of which I don't know whether it would be approved. Specifically the sum_image_pixels in [integral_image](src/integral_image.rs). The issue is that when changing to use GenericImageView, the compiler didn't like accessing the returned array stating it could index into the variable. I changed (at least temporarily) to return a Pixel rather than the array.

While I don't expect this specific change to be accepted, if there are any suggestions as to how best to deal with it it would be appreciated. I also still need to write tests for multi-channel images.

### DIFF
--- a/examples/template_matching.rs
+++ b/examples/template_matching.rs
@@ -98,7 +98,7 @@ fn run_match_template(
     method: MatchTemplateMethod,
 ) -> RgbImage {
     // Match the template and convert to u8 depth to display
-    let result = match_template(&image, &template, method);
+    let result = match_template(image, template, method);
     let result_scaled = convert_to_gray_image(&result);
 
     // Pad the result to the same size as the input image, to make them easier to compare
--- a/src/integral_image.rs
+++ b/src/integral_image.rs
@@ -4,7 +4,7 @@
 use crate::definitions::Image;
 use crate::map::{ChannelMap, WithChannel};
 use image::{GenericImageView, GrayImage, Luma, Pixel, Primitive, Rgb, Rgba};
-use std::ops::AddAssign;
+use num::traits::*;
 
 /// Computes the 2d running sum of an image. Channels are summed independently.
 ///
@@ -50,7 +50,8 @@ use std::ops::AddAssign;
 pub fn integral_image<P, T>(image: &Image<P>) -> Image<ChannelMap<P, T>>
 where
     P: Pixel<Subpixel = u8> + WithChannel<T> + 'static,
-    T: From<u8> + Primitive + AddAssign + 'static,
+    P::Subpixel: NumCast,
+    T: Primitive + NumAssign + 'static,
 {
     integral_image_impl(image, false)
 }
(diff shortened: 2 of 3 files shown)


theotherphil:
Do you need the ability to support any GenericImageView here, or could you get away with just implementing this for Image<T> for T: Pixel + 'static? The latter is obviously less flexible, but makes the code a lot simpler. And potentially allows for faster code (several functions in this library started life operating on any generic image and were then switched to requiring a fixed image type to make them simpler and faster).

JonBoyleCoding:
I was originally trying to work with sub-images; however, perhaps as you say it would be simpler to do with expanding GreyImage to Image/ImageBuffer. I'll get this done soon and send a new pull request.

Thanks for your advice.

JonBoyleCoding:
I understand this isn't mergable yet, but thought it best way to ask any thoughts. I reverted integral_image and with some minor changes to turn things into Image, but I still have the same issue with ArrayData::Datatype. Do you have any thoughts on this?

theotherphil:
Thanks. I’ll have a look on the weekend.

theotherphil:
I didn’t get around to looking at this after all. I’m definitely interested in having this functionality in the library, but have very little time available for working on this crate at the moment so I’m afraid I’m not doing to do a great job of responding to questions or PRs that require much effort from me.

JonBoyleCoding:
I understand, thank you for responding!

I've been thinking about it as well, and I think one good way to go about this if concerned about api changes is to have sum_image_pixels return a vec of the data rather than a straight array. While this would be a slight change, I think it would be quite minimal in that I don't think it would conflict with anyone's code bases. I'll implement a version of this and push to this Pull Request, then hopefully it will be a lot less time on your part to see if you're ok with the change and merge.

JonBoyleCoding:
I have it all implemented I think, I changed a number of traits to be based on the Num crate as it made the casting a whole lot easier. I notice out of the CI build that beta/nightly have succeeded, but not stable, although I can't see any of the issues on the logs of the stable.

I've added in tests as well, and fixed the documentation of the sum_image_pixels function for the new output. Hopefully it's in good condition for you to accept, but let me know if it isn't.

JonBoyleCoding:
Figured out the stable compile issue, I forgot to format the code!

ChemicalXandco:
Is there anything preventing this from being merged?

JonBoyleCoding:
I would imagine as this PR was from over 2 years ago, that it might need to be updated. I'll try and have a look over the next few days to see.

ripytide:
Since this PR is quite old now and when I attempted to rebase it I got quite a few conflicts, would it be a good idea to close it? I think closing is the better alternative to simply leaving it open for many more years which makes it harder to focus on new PRs due to buildup.

theotherphil:
Sounds sensible.

PR #427, opened 2020-08-29
Update stats function to take GenericImageView

tinou98:
All function only need access to a `GenericImageView`, not a `GenericImage`. This should avoid a call to `to_image`.

This is not a breaking change because all `GenericImage` must implement `GenericImageView`.


theotherphil:
Thanks!

PR #428, opened 2020-09-01
Improved draw_polygon fn

lazareviczoran:
I've added `draw_polygon/draw_polygon_mut` fns, in which I've reused most of the `draw_convex_polygon` and modified it a bit in order to be able to draw concave polygons as well.

Also, I've noticed a bug with drawing the elements that are partially outside of the canvas, and fixed that as well.
Screen Shot 2020-09-01 at 3 59 25 PM

I added a test for this (same as the draw_convex_polygon test, with an aditional concave shape), so you can compare it with the `polygon.png`.


theotherphil:
Thanks!

Don't bother deprecating the old functions - just delete them and I'll mark the release as backwards-incompatible. This should hopefully also give a more readable diff!

src/drawing/polygon.rs
```diff
@@ -129,3 +131,104 @@ where
         draw_line_segment_mut(canvas, start, end, color);
     }
 }
+
+/// Draws as much of a filled concave polygon as lies within image bounds. The provided
```
theotherphil, line 135: Or non-concave!

theotherphil:
Sorry for the long delay in reviewing this - looking now!

src/drawing/polygon.rs
```diff
@@ -58,38 +53,41 @@ where
         );
     }
 
-    let mut y_min = i32::MAX;
-    let mut y_max = i32::MIN;
-    for p in poly {
-        y_min = min(y_min, p.y);
-        y_max = max(y_max, p.y);
-    }
+    let (mut y_min, mut y_max) = poly
```
theotherphil, line 56: Nit: I find this much harder to read at a glance than the code it replaces. Don't feel strongly though.

src/drawing/polygon.rs
```diff
@@ -99,33 +97,26 @@ where
...
             }
-            i += 2;
-        }
+        });
 
         intersections.clear();
     }
 
-    for edge in &edges {
+    edges.iter().for_each(|edge| {
```
theotherphil, line 117: Nit: this is also a non-functional tweak that doesn't help with code readability. In this case the old code is also more idiomatic.

theotherphil:
I've not thought through the new algorithm, but my first thought is whether it correctly handles polygons such as

image

Could you add a test that covers something like this?

lazareviczoran:
> I've not thought through the new algorithm, but my first thought is whether it correctly handles polygons such as
> 
> image
> 
> Could you add a test that covers something like this?

Sure, I've added a new test for this shape, if you have more examples, send me a picture, and I can add them to the test (or feel free to add them yourself)

theotherphil:
Ah! I coloured that confusingly. White is the background and black is the polygon. I wanted to check that this works fine if there are background pixels that have polygon pixels between them and any point on the frame of the image.

src/drawing/polygon.rs
```diff
@@ -71,25 +66,29 @@ where
...
+                    // Need to handle horizontal lines specially
                     intersections.push(p0.x);
                     intersections.push(p1.x);
+                } else if p0.y == y || p1.y == y {
+                    if p1.y > y {
+                        intersections.push(p0.x);
+                    }
+                    if p0.y > y {
+                        intersections.push(p1.x);
+                    }
```
lazareviczoran, line 91: This is the main difference in the algorithm (handling the case when `y` is at the end of an edge).
There are 3 cases when calculating the intersection with `y` being in the end point of an edge:
- when both edges that end in `y` are above `y` (we will count them both)
image

- when both edges that end in `y` are bellow `y` (we will ignore them both)
image

- when edges that end in `y` have their ends on the opposite sides compared to `y` (we will count just one)
image

The previous algorithm was counting both ends always, which messed up the drawing in case of a concave polygon

lazareviczoran:
> Ah! I coloured that confusingly. White is the background and black is the polygon. I wanted to check that this works fine if there are background pixels that have polygon pixels between them and any point on the frame of the image.
> 
> If that works fine then this is good to go. Thanks!

I've updated the test, take a look if that's what you meant.

theotherphil:
Thanks, this is a big improvement to these functions.

The test you added isn't quite what I had in mind, but I've checked it locally and it works fine so I'll tweak the test in a follow up PR.

Issue #317, opened 2019-03-23
Add contour finding algorithm 


MichaelMauderer:
Something like https://docs.opencv.org/3.4/d3/dc0/group__imgproc__shape.html#ga95f5b48d01abc7c2e0732db24689837b

mikemoraned:
FYI, I just opened https://github.com/image-rs/imageproc/issues/406 which is similar to what you'd like. Want to work on this together in some way?

MichaelMauderer:
I'd love to, but I don't really have the spare capacity for it at the moment.
(I went the alternative route and used [opencv-rust](https://github.com/twistedfall/opencv-rust).)

PR #430, opened 2020-10-05
Added a contours finding function

lazareviczoran:
Implementing a contours tracing algorithm (https://www.semanticscholar.org/paper/Topological-structural-analysis-of-digitized-binary-Suzuki-Abe/cf021db5e811fd5b67ee3aa4db0a6a0351d276d2)

Within the `find_contours_with_thresh` fn, I'm handling the coordinates as `usize` (since its values are in range [0..width) and [0..height), and it is easier to use for indexing the image_values matrix), with occasional conversions to `i32` for an easier neighbour position calculation and out of bounds check, but at the end, since the function is written with generics (with a limitation on the numeric types using the `num` crate), when storing the contour points the coordinates are converted to that generic type.

Also, I've moved the `Point<T>` struct from `src/drawing/polygon.rs` to `src/definitions.rs`, since it is a base struct that can be reused on other places as well (like https://github.com/image-rs/imageproc/blob/master/src/drawing/line.rs#L188, or https://github.com/image-rs/imageproc/blob/master/src/drawing/bezier.rs#L28)

Issue: https://github.com/image-rs/imageproc/issues/317

### DIFF
--- a/src/contours.rs
+++ b/src/contours.rs
@@ -1,0 +1,187 @@
+//! Functions for detecting contours of polygons in an image and approximating
+//! polygon from set of points.
+
+use crate::definitions::Point;
+use image::GrayImage;
+use num::{cast, Num, NumCast};
+use std::collections::VecDeque;
+
+/// Contour struct containing its points, is_outer flag to determine whether the
+/// contour is an outer border or hole border, and the parent option (the index
+/// of the parent contour in the contours vec)
+#[derive(Debug)]
+pub struct Contour<T: Num + NumCast + Copy + PartialEq + Eq> {
+  /// All the points on the contour
+  pub points: Vec<Point<T>>,
+  /// Flag to determine whether the contour is an outer border or hole border
+  pub is_outer: bool,
+  /// the index of the parent contour in the contours vec, or None if the
+  /// contour is the outermost contour in the image
+  pub parent: Option<usize>,
+}
+impl<T: Num + NumCast + Copy + PartialEq + Eq> Contour<T> {
+  /// Construct a contour.
+  pub fn new(points: Vec<Point<T>>, is_outer: bool, parent: Option<usize>) -> Self {
+    Contour {
+      points,
+      is_outer,
+      parent,
+    }
+  }
+}
+
+/// Finds all the points on the contours on the provided image.
+/// Handles all non-zero pixels as 1.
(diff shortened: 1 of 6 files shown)


src/contours.rs
```diff
@@ -0,0 +1,193 @@
...
+use image::GrayImage;
+use num::{cast, Num, NumCast};
+use std::collections::VecDeque;
+
+/// Contour struct containing its points, is_outer flag to determine whether the
+/// contour is an outer border or hole border, and the parent option (the index
+/// of the parent contour in the contours vec)
+#[derive(Debug)]
+pub struct Contour<T: Num + NumCast + Copy + PartialEq + Eq> {
+    /// All the points on the contour
```
theotherphil, line 14: Nit: doc text lines should end with a '.' as otherwise it looks odd in the rendered docs.

theotherphil:
cc @mikemoraned who also has an implementation of contour finding that they were thinking of adding to this library.

src/contours.rs
```diff
@@ -0,0 +1,193 @@
...
+        (-1, 1),  // sw
+    ]);
+    let mut x = 0;
+    let mut y = 0;
+    let last_pixel = (width - 1, height - 1);
+
+    let mut contours: Vec<Contour<T>> = Vec::new();
+    let mut skip_tracing;
+    let mut nbd = 1;
+    let mut lnbd = 1;
```
theotherphil, line 81: Could you give clearer names to these two values.

tests/regression.rs
```diff
@@ -622,3 +622,107 @@ fn test_hough_line_detection() {
         assert_pixels_eq!(lines_image, truth);
     }
 }
+
```
theotherphil, line 625: The tests in this file all compare generated images to images under tests/data/truth.

Tests that don't do I/O should go in the same file as the code they're testing.

src/contours.rs
lazareviczoran: Sure, I've used the notations from the [paper](https://www.semanticscholar.org/paper/Topological-structural-analysis-of-digitized-binary-Suzuki-Abe/cf021db5e811fd5b67ee3aa4db0a6a0351d276d2), but it makes sense to give them clearer names.

theotherphil:
Looks good to me, but needs merge conflicts resolving now that I've merged your previous PR.

theotherphil:
Thanks!

PR #434, opened 2020-10-10
better use of methods provided by std

TianyiShi2001:
## Summary

- change `let (s, c) = (deg.sin(), deg.cos)` to `let (s, c) = deg.sin_cos()`
- use the `f32::to_radians()` provided by std instead of reinventing the wheel


src/hough.rs
```diff
@@ -159,8 +159,7 @@ fn intersection_points(
     }
 
     let theta = degrees_to_radians(m);
```
theotherphil, line 161: Missed one! This is what’s causing the Travis build to fail.

theotherphil:
Thanks!

PR #436, opened 2020-10-10
use type alias

TianyiShi2001:
According to Wikipedia, these 3x3 matrices are termed "affine matrix" https://en.wikipedia.org/wiki/Affine_transformation#Image_transformation


TianyiShi2001:
See also #437 about the type alias

theotherphil:
Sorry, I'm going to close this one without merging. I don't think this is an improvement.

PR #438, opened 2020-10-10
cleaner syntax

TianyiShi2001:
`let (a0, a1, a2) = (aa[0], aa[1], aa[2])` can be rewritten as `let [a0, a1, a2] = aa` as long as the size is fixed


theotherphil:
Nice!

PR #444, opened 2020-10-14
Fixed contour finding for single pixel elements, fixed parent assignment

### LINKED ISSUE #443: Find contours skips single pixel contours
On images like 
Screen Shot 2020-10-14 at 10 41 20 AM
the `find_contours` fn, skips detecting single pixel elements.

lazareviczoran:
Fixes #443.

I've modified the algorithm to not skip single pixel elements. 
Besides that, I've fixed the parent assignment for a contour (wasn't working correctly when multiple object are detected on the same height) and fixed the contour labeling.

Basically, I've misunderstood parts of the algorithm from the [paper](https://www.semanticscholar.org/paper/Topological-structural-analysis-of-digitized-binary-Suzuki-Abe/cf021db5e811fd5b67ee3aa4db0a6a0351d276d2), which I figured out when running the algorithm on a more complex image, and fixed them to work properly.

I've added a test for image (the image is zoomed in, the right element is a single pixel)
Screen Shot 2020-10-14 at 10 41 20 AM

which should result in
Screen Shot 2020-10-14 at 10 54 49 AM
as stated in the paper.

### DIFF
--- a/src/contours.rs
+++ b/src/contours.rs
@@ -86,30 +86,35 @@ pub fn find_contours_with_thresh<T: Num + NumCast + Copy + PartialEq + Eq>(
     while (x, y) != last_pixel {
         if image_values[x][y] != 0 {
             skip_tracing = false;
+            let is_outer;
             if image_values[x][y] == 1 && x > 0 && image_values[x - 1][y] == 0 {
                 curr_border_num += 1;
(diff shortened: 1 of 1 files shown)


theotherphil:
Thanks!

Issue #98, opened 2016-01-11
How best to add parallelism?


theotherphil:
Lots of the functions in this library are embarrassingly parallel. How should we parallelise them? Ideally with as little impact on the function bodies as possible. Image-chunk-iterators plus https://github.com/nikomatsakis/rayon?

tafia:
On the *get* side, for most functions we could simply change `for y in 0..height` into a rayon equivalent.
IMHO, pb is on *put* side, ie sharing mutable `ImageBuffer`:
- either we stop returning `ImageBuffer` and return iterators
  - pros: 
    - we could probably avoid a lot of `unsafe_put_pixel`
    - perhaps (?) we could chain several fns in an efficient way?
  - cons: 
    - much more complicated
    - we don't reserve memory upfront
- or we add a `chunk_mut` like method on ImageBuffer itself

theotherphil:
Yeah, I've thought (very little) about lazy image operations before, but haven't looked into how feasible this is. There are several places where returning iterators could be useful in order to avoid allocating intermediates at all, so the second con could be a positive.

Adding the ability to get mutable references to disjoint sections of an image sounds like the most straightforward approach, but I'm really not qualified to give much useful input on this issue.

nicokoch:
What about providing implementations that run on the GPU, i.e. with openCL or OpenGL compute shaders?

theotherphil:
That would be great to have. Is there anything in the rust ecosystem already which would help with this? All I can see from a quick Google is https://github.com/luqmana/rust-opencl.

tafia:
The work on vulkan might probably help, even if it means going into the
unknown.
On 26 Feb 2016 13:48, "theotherphil" <notifications@github.com> wrote:

That would be great to have. Is there anything in the rust ecosystem
already which would help with this? All I can see from a quick Google is
https://github.com/luqmana/rust-opencl.

—
Reply to this email directly or view it on GitHub
<https://github.com/PistonDevelopers/imageproc/issues/98#issuecomment-189123392>
.

theotherphil:
A totally different, possibly terrible, idea is to write a rust front-end for http://halide-lang.org/.

nicokoch:
https://github.com/tomaka/glium is a great OpenGL wrapper for rust (I contribute there a lot). It has support for OpenGL compute shaders, which are kind of an alternative to OpenCL.

[...]

PR #446, opened 2020-11-09
Map iter

### LINKED ISSUE #429: Parallelize map_colors/map_pixels/map_*
I am trying to find an _easy_ way to parallelize [`map_colors`](https://docs.rs/imageproc/0.21.0/imageproc/map/fn.map_colors.html). Does anyone has a solution for that?

From an user perspective, I believe that the easiest way would be to provide an iterator over colors/pixels/subpixels/whatever and then the user could wrap it with [rayon](https://docs.rs/rayon/1.4.0/rayon/)'s `par_iter`. It may be related to #395, but it hasn't moved lately.

carrascomj:
Closes #429 
Related to #98 
### Description
Functions under `src/map.rs` are subject to parallelization. Thus, we can use [rayon](https://github.com/rayon-rs/rayon) to distribute them across threads.

### Implementation
- Refactored everything in `src/map.rs` from a loop to use Iterators from [Pixels](https://docs.rs/image/0.23.11/image/struct.Pixels.html). This provides a suitable structure to parallelize it.
- Under the `rayon` feature, compile a version of `map_colors()` that uses `par_bridge`.

### TODO
Before continuing with the rest of the functions, I'd like to know if this is a good direction:
- [ ] Is there a better way to flush the pixels into the ImageBuffer? Currently, the pixels are parallelized and transformed into a `Vec`, to then gather them (single thread) into a ImageBuffer with `unsafe_put_pixel`. Alternatively, a channel could be considered like in the [rust cookbook](https://rust-lang-nursery.github.io/rust-cookbook/concurrency/threads.html#draw-fractal-dispatching-work-to-a-thread-pool).
- [ ] _Implemented [in this other branch](https://github.com/carrascomj/imageproc/blob/pixeliter/src/map.rs)_: should an additional custom struct that implements iterator over `GenericImage` be introduced instead? That would allow for using `unsafe_get_pixel` (like it is on master), but I am unsure if that would have a strong justification. [The implementation of `Iterator` of `Pixels`](https://docs.rs/image/0.23.11/src/image/image.rs.html#566) is safe and does bound checking.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1,6 +1,8 @@
 //! Functions for mapping over pixels, colors or subpixels of images.
 
 use image::{Bgr, Bgra, GenericImage, ImageBuffer, Luma, LumaA, Pixel, Primitive, Rgb, Rgba};
+#[cfg(feature = "rayon")]
+use rayon::prelude::*;
 
 use crate::definitions::Image;
 
@@ -62,7 +64,6 @@ where
 {
     type Pixel = LumaA<U>;
 }
-
 
 /// Applies `f` to each subpixel of the input image.
 ///
@@ -134,6 +135,7 @@ where
 ///     rgb);
 /// # }
 /// ```
+#[cfg(not(feature = "rayon"))]
 pub fn map_colors<I, P, Q, F>(image: &I, f: F) -> Image<Q>
 where
     I: GenericImage<Pixel = P>,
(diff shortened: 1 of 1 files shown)


carrascomj:
I did some benchmarks with [`draw_vertical_seams()`](https://github.com/image-rs/imageproc/blob/master/src/seam_carving.rs#L162) and it turned out that the different implementations proposed (also tried with a crossbeam channel and confirmed that it was using all processors) with rayon are orders of magnitude slower than the sequential one.

PR #448, opened 2020-11-12
Modified convex_hull fn to handle coordinates with negative values

lazareviczoran:
The `convex_hull` fn param -> `points_slice: &[Point<T>]` is a set of points, there is no specific limit (like `x = {0,width}` and `y = {0,height}` when the param is an image).
Since in the original implementation I was converting it to `usize` it was failing when the generic type T included negative values.
Actually, that conversion can be avoided by including `Ord` ~~and `Bounded`~~ trait~~s~~ to the generic type definition.


theotherphil (approved):

theotherphil:
Thanks!

PR #453, opened 2020-11-28
Add function to get rendered size of text

dmarcuse:
This PR adds a simple `text_size` function that returns the width and height of text that would be rendered with a given font and scale. This is useful to avoid having to manually call `rusttype`'s layout functions, and ensure the results match the actual text rendered with `draw_text`/`draw_text_mut`.

The `drawing` module may not be the best place for this function since it doesn't actually draw anything - I'd be happy to move it to another module if appropriate.

It would probably be good to additionally update `draw_text` or `draw_text_mut` to also return rendered text size, but this would be a breaking change as far as I'm aware.

### DIFF
--- a/examples/font.rs
+++ b/examples/font.rs
@@ -1,7 +1,7 @@
 //! An example of drawing text. Writes to the user-provided target file.
 
 use image::{Rgb, RgbImage};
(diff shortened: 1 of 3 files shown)


theotherphil:
Thanks!

PR #456, opened 2021-01-04
faster combination of common transformation classes (translation & affine)

TianyiShi2001:
It is not uncommon to combine only translations and/or affine transformations (without combining projections). These combinations can be computed faster than using the generic 3x3 matrix multiplication. The speed improvement is demonstrated in benchmarks on commit https://github.com/image-rs/imageproc/pull/456/commits/74b57890cb84cf2dee5d666325cb1248fa1b4fa4 :

Running `cargo bench translationx` on commit https://github.com/image-rs/imageproc/pull/456/commits/74b57890cb84cf2dee5d666325cb1248fa1b4fa4 :

```
running 4 tests
test tests::translationxrotation         ... bench:   3,966,842 ns/iter (+/- 318,200)
test tests::translationxrotation_fast    ... bench:   2,993,829 ns/iter (+/- 88,294)
test tests::translationxtranslation      ... bench:   3,945,007 ns/iter (+/- 302,402)
test tests::translationxtranslation_fast ... bench:   3,009,790 ns/iter (+/- 243,091)
```


theotherphil:
I’m not convinced that the extra complexity here is worth it - do you have a use case in mind where people would be combining enough distinct transformations that the cost of a 3x3 matrix multiplication would be noticeable?

PR #457, opened 2021-01-04
add `and_then` method for `Projection`

TianyiShi2001:
It might be more intuitive for people who are not familiar with linear algebra.
This also benefit from method name hints in IDEs/text editors.


theotherphil:
Thanks!

PR #463, opened 2021-04-04
Fix assert! macro string literal warning

ireina7:
This PR fixed macro `assert!` usage warning:
```
warning: panic message is not a string literal
  --> src/filter/mod.rs:94:13
   |
94 | /             format!(
95 | |                 "Invalid kernel len: expecting {}, found {}",
96 | |                 width * height,
97 | |                 data.len()
98 | |             )
   | |_____________^
   |
   = note: `#[warn(non_fmt_panic)]` on by default
   = note: this is no longer accepted in Rust 2021
   = note: this warning originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

warning: 1 warning emitted
```
The warning has been fixed in this PR by using *string literal* instead.


theotherphil:
Thanks!

PR #477, opened 2021-11-26
Bump dependencies

paolobarbolini:


theotherphil:
Thanks. Have you run the tests locally? Looks like travis integration on PRs has stopped working at some point - I’ll have to fix that later.

paolobarbolini:
I did and they all pass. Running cargo test on stable though complains with `#![cfg_attr(test, feature(test))]`, and cargo doc throws warnings about broken doc links, but I guess it was already like so before.

theotherphil:
Thanks!

PR #480, opened 2021-12-13
Update text.rs

gtimoshaz:
Fixed misleading information in docstring


theotherphil:
Thanks. This function also has an incorrect signature - the input image reference does not need to be mutable. I’d be happy to accept a PR fixing this if you’re willing!

Issue #482, opened 2022-01-04
Improve the documentation/explanation of drawing functions


alexwlchan:
In particular looking at https://docs.rs/imageproc/0.22.0/imageproc/drawing/index.html, a few things jump out at me:

* The short descriptions are somewhat inconsistent, e.g. whether they're described as "in place" or not (and some of them sound wrong, e.g. ~is `draw_text` drawing text in place?~ fixed in #480)
* Some of the information seems like it would be better pushed into the detailed description on the per-function pages, e.g. the exact ellipse-drawing algorithm
* When should I use the in-place vs the not in-place algorithm? I can make an educated guess, but it'd be good to know what the implications of either approach are.

[Context: I'm writing a blog post that mentions imageproc, and I'm having to handwave about why I'm using the `_mut` methods.]

PR #484, opened 2022-01-09
Tidy up the description of functions in the `drawing` module

alexwlchan:
* I've edited the descriptions so they just get a one-line description in the module overview, making it easier to skim
* I've tweaked the wording so every function is consistent about how it describes, e.g. a filled shape vs the outline of a shape
* I've added consistent wording for "in place" vs "on a new copy of the image", and added a sentence clarifying what the difference between the two variants of function are
* I've fixed a broken link to a description of the midpoint ellipse drawing algorithm

Before:

Screenshot 2022-01-09 at 13 47 12

After:

Screenshot 2022-01-09 at 13 47 15

Closes https://github.com/image-rs/imageproc/issues/482

### DIFF
--- a/src/drawing/bezier.rs
+++ b/src/drawing/bezier.rs
@@ -5,7 +5,9 @@ use image::{GenericImage, ImageBuffer};
 use std::f32;
 use std::i32;
 
-/// Draws as much of a cubic bezier curve as lies within image bounds.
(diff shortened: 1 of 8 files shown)


theotherphil:
A big improvement, thanks!

PR #487, opened 2022-02-10
Add must_use attribute to non-mutating functions.

surban:
This annotates functions that do not modify the original image.


theotherphil:
Thanks!

Issue #498, opened 2022-05-21
add anti-aliased options for drawing functions


zainthemaynnn:
I would love to have anti-aliased versions of the other functions under `drawing` and not just raw line segments. as far as I know this is just a matter of calling the existing functions with `draw_antialiased_line_segment_mut` instead of `draw_line_segment_mut`? I can probably add it myself.

Issue #496, opened 2022-05-01
Feature request: `draw_filled_polygon[_mut]` (drawing polygon with fill but no stroke/edges)


finnbear:
As I understand it, the `drawing` package contains `draw_polygon[_mut]` which draw a polygon with fill and edges. For my use-case, I want the fill but not the edges. I don't see a way to do that, as the existing function ends [by unconditionally drawing the edges](https://docs.rs/imageproc/latest/src/imageproc/drawing/polygon.rs.html#108-111).

I would benefit from something like `draw_filled_rect[_mut]`, that draws the fill and not the edges.

PR #499, opened 2022-05-22
add draw_antialiased_polygon and draw_unbordered_polygon

zainthemaynnn:
this solves (#498) and while I was at it, (#496)

### DIFF
--- a/src/drawing/mod.rs
+++ b/src/drawing/mod.rs
@@ -25,7 +25,11 @@ pub use self::line::{
 };
 
 mod polygon;
-pub use self::polygon::{draw_polygon, draw_polygon_mut};
+pub use self::polygon::{
+    draw_polygon, draw_polygon_mut,
+    draw_unbordered_polygon, draw_unbordered_polygon_mut,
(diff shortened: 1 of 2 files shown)


theotherphil:
Thanks for your PR. Please can you add some regression tests for your new functions that show them working.

https://github.com/image-rs/imageproc/blob/6a7adf4bfc75e506452c43681ab9b9a90ecb4081/tests/regression.rs#L413

and

https://github.com/image-rs/imageproc/blob/6a7adf4bfc75e506452c43681ab9b9a90ecb4081/tests/regression.rs#L477

are the tests that exist for the current polygon functions. There is a comment at the top of that file explaining how to (re)generate the “truth” images.

zainthemaynnn:
it's a good thing, too, because it looks like the unbordered ones didn't actually work. I guess it's not as easy as I thought since it's impossible to draw the polygon without the bresenham exactly unless you actually run the bresenham. so that was silly.

anyways, I have added a test for the anti-aliased function with the same shapes from the aliased truth image.

theotherphil:
Thanks!

PR #500, opened 2022-05-22
Optimize interpolate_nearest

mat-1:
Right now interpolate_nearest does .round() to get the positions, but adding 0.5 and casting is faster. I didn't benchmark interpolate_nearest specifically, but my program that makes heavy use of it became around 3 times faster with this change.

### DIFF
--- a/src/geometric_transformations.rs
+++ b/src/geometric_transformations.rs
@@ -710,14 +710,18 @@ where
 
 #[inline(always)]
 fn interpolate_nearest<P: Pixel>(image: &Image<P>, x: f32, y: f32, default: P) -> P {
-    let rx = (x + if x.is_sign_positive() { 0.5 } else { -0.5 }) as i32;
-    let ry = (y + if y.is_sign_positive() { 0.5 } else { -0.5 }) as i32;
-
     let (width, height) = image.dimensions();
-    if rx < 0 || ry < 0 || rx as u32 >= width || ry as u32 >= height {
+    if x < -0.5 || y < -0.5 {
+        return default;
+    }
+
+    let rx = (x + 0.5) as u32;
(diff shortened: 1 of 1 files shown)


MattTheCuber (approved):

src/geometric_transformations.rs
```diff
@@ -710,11 +710,11 @@ where
...
 #[inline(always)]
 fn interpolate_nearest<P: Pixel>(image: &Image<P>, x: f32, y: f32, default: P) -> P {
-    let rx = x.round();
-    let ry = y.round();
+    let rx = (x + if x.is_sign_positive() { 0.5 } else { -0.5 }) as i32;
+    let ry = (y + if y.is_sign_positive() { 0.5 } else { -0.5 }) as i32;
 
     let (width, height) = image.dimensions();
-    if rx < 0f32 || rx >= width as f32 || ry < 0f32 || ry >= height as f32 {
+    if rx < 0 || rx >= width as i32 || ry < 0 || ry >= height as i32 {
```
fintelia, line 717: What if width or height is larger than i32::MAX?

mat-1: Oh thanks for pointing that out I didn't consider that before. I just fixed it by casting width/height to u32 instead of casting rx/ry to i32 and reordering it to make sure they aren't less than 0 first.

fintelia (approved):
This function is going to treat NaNs as zero. Not sure if that is a problem, but it is something to be aware of

mat-1:
> This function is going to treat NaNs as zero. Not sure if that is a problem, but it is something to be aware of

I didn't even know f32::NAN was a thing, but the old code has that same behavior of treating NaNs as 0 so I think it's fine.

theotherphil:
Thanks!

Issue #502, opened 2022-07-01
draw_hollow_polygon(_mut)


dan-jfisher:
Not really a feature request because I have already written it in my own project.  

I'm just creating an issue so that I can add it here as well.

PR #503, opened 2022-07-01
add draw_hollow_polygon and draw_hollow_polygon_mut

dan-jfisher:
#502 

This PR adds functions to draw a series of line segments to form a hollow polygon, similar to draw_hollow_rect.

In practice it means you can draw tthings like the output of min_area_rect more easily.


theotherphil:
Thanks!

PR #511, opened 2022-09-22
add floodfill, floodfill_mut, neighbors, and neighbor_indices

mcaveniathor:
Added functions to yield iterators to neighboring pixel indices or references to the pixels, in-place floodfill, and floodfill.

### DIFF
--- a/src/point.rs
+++ b/src/point.rs
@@ -1,4 +1,5 @@
 //! A 2d point type.
+use crate::definitions::Position;
 
 use num::{Num, NumCast};
 use std::ops::{Add, AddAssign, Sub, SubAssign};
@@ -16,6 +17,12 @@ impl<T> Point<T> {
     /// Construct a point at (x, y).
     pub fn new(x: T, y: T) -> Point<T> {
         Point::<T> { x, y }
+    }
+}
+
+impl<T: Num> From<Point<T>> for (T, T) {
+    fn from(point: Point<T>) -> Self {
+        (point.x, point.y)
     }
 }
 
@@ -59,6 +66,17 @@ impl<T: NumCast> Point<T> {
     pub(crate) fn to_i32(&self) -> Point<i32> {
         Point::new(self.x.to_i32().unwrap(), self.y.to_i32().unwrap())
     }
+}
+
+
+impl Position for Point<u32> {
+    fn x(&self) -> u32 { self.x }
+    fn y(&self) -> u32 { self.y }
+}
+
+impl Position for Point<usize> {
+    fn x(&self) -> u32 { self.x as u32 }
+    fn y(&self) -> u32 { self.y as u32 }
 }
 
 /// Returns the Euclidean distance between two points.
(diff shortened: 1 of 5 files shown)


mcaveniathor:
I have implemented floodfill, floodfill_mut, as well as created the _connectivity_ module and its utilities for working with pixels' neighbors.

src/point.rs
```diff
@@ -61,6 +68,17 @@ impl<T: NumCast> Point<T> {
...
+
+impl Position for Point<u32> {
+    fn x(&self) -> u32 { self.x }
+    fn y(&self) -> u32 { self.y }
+}
+
+impl Position for Point<usize> {
+    fn x(&self) -> u32 { self.x as u32 }
+    fn y(&self) -> u32 { self.y as u32 }
+}
```
ripytide, line 80: It's probably not a great idea to implement this trait using casts, it could cause some very hard to track down bugs. It might be a better idea to do this conversion manually where needed.

src/region_labelling.rs
```diff
@@ -241,6 +238,102 @@ where
...
+
+
+
+
+
+
+
+
+
+
```
ripytide, line 336: Might need to run `cargo fmt`

src/region_labelling.rs
```diff
@@ -241,6 +238,102 @@ where
...
+where
+    P: Pixel+PartialEq,
+    T: Position
+{
+    let seed_point = (seed_point.x(), seed_point.y());
+    let (width, height) = image.dimensions();
+    let old_color = image[seed_point]; // connected pixels matching this color will be filled in with the fill_color
+    if old_color == fill_color || seed_point.0 >= width || seed_point.1 >= height {
+        return;
+    }
```
ripytide, line 285: clippy warns about an unneeded return statement

src/region_labelling.rs
```diff
@@ -241,6 +238,102 @@ where
...
+    else {
+        let mut frontier = Vec::with_capacity(50); // preallocation size chosen arbitrarily as the required size varies
+        frontier.push(seed_point);
+        image[seed_point] = fill_color;
+        let connectivity = Connectivity::Four;
+        while !frontier.is_empty() {
+            let q = frontier.pop().unwrap(); // unwrap will not panic here as we have checked that frontier is not empty
```
ripytide, line 292: You could re-write this as:
```rust
        while let Some(q) = frontier.pop() {
```

src/region_labelling.rs
```diff
@@ -241,6 +238,102 @@ where
...
+                    image[neighbor] = fill_color;
+                }
+            }
+        }
+
+    }
+}
+
+/// Creates a clone of the input image and performs an in-place floodfill operation on 
+/// the clone using [floodfill_mut]
```
ripytide, line 307: I think references to functions in doc-comments should also use backticks like: ``[`floodfill_mut()`]``

ripytide (approved):
I like that connectivity is defined into it's own file. The PR lacks some regular tests but it is well documented and has doc-examples.

theotherphil:
This functionality is definitely nice to have. I suspect that this version will be far slower than it should be though - we're both allocating a vector and performing bounds checks for every pixel in the image.

PR #514, opened 2023-02-13
Sgx nostd compat

n-prat:


theotherphil:
Sorry for the long delay in getting to this PR. I don't think adding SGX support is worth the complexity it adds.

n-prat:
No worry.

It was more about `no_std` than SGX, I should have been more clear.
To be fair, it is not really complex : it touches a lot of files, but it is mostly touching imports and renaming a few things.

PR #515, opened 2023-02-22
fix 100% cpu usage on Linux

alxpettit:
Confirmed working on Arch Linux, with far less wasted CPU resources. Also contains a slight refactoring to replace the duplicate code path with a single one, using a `|` to check multiple possibilities in one path, in the `match`.


theotherphil:
Thanks!

alxpettit:
Welcome ^w^

PR #517, opened 2023-04-17
add a general nxm convolution filter (a generalization of filter3x3)

vsuryamurthy:
This PR is a generalization of the filter3x3 and extends to odd shaped kernels of dimensions nxm.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -356,6 +356,21 @@ where
     kernel.filter(image, |channel, acc| *channel = S::clamp(acc))
 }
 
+/// Returns 2d correlation of an image with a nxm row-major odd shaped kernel. Intermediate calculations are
+/// performed at type K, and the results clamped to subpixel type S. Pads by continuity.
+#[must_use = "the function does not modify the original image"]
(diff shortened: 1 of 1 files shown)


cospectrum:
Maybe we can just call it `filter`? 🤔

cospectrum:
Or we can expose it to users as a Kernel method. `clamped_filter`, for example.

vsuryamurthy:
Sure, either works. What would be the prefernce?

cospectrum:
> Sure, either works. What would be the prefernce?

Let's ask the owner @theotherphil. Do you have any preferences or maybe some criticism of this pr?

cospectrum:
The strange thing is that this only works for odd sizes. I have an implementation of 2d convolution in Rust by matrix multiplication (compatible with [torch](https://pytorch.org/docs/stable/generated/torch.nn.functional.conv2d.html)), and the kernels are of any size.

theotherphil:
Thanks for the PR. I’m not sure about the filternxm function - filter3x3 exists mainly because it initially had a special cased implementation. I’d accept a PR to add a clamped_filter function to Kernel.

PR #521, opened 2023-05-09
Make `Contour` clonable.

de-vri-es:
I want to clone some custom structs that contain contours. And I don't see any reason why `Contour` should not implement `Clone` :)


de-vri-es:
@theotherphil: sorry to do a personal mention, but do you have time to review this change?

theotherphil:
Thanks!

Issue #524, opened 2023-06-01
Interest for a parallel version of match_template?


stephanemagnenat:
Hello,

For a personal project, I wrote a parallel version of `match_template`, keeping essentially the same algorithm but using a Rayon parallel iterator for the y coordinate and collecting the result.

If there is interest here, I'm happy to do a PR with my changes. It would be good to know whether this should be an addition or a replacement, and if some tests are needed.

kind regards

stephanemagnenat:
I'll do a benchmark and report the results tomorrow, but I would say, close to the number of cores.

stephanemagnenat:
The PR is ready: #527.

I went for simplicity, if we are all happy about this new feature, it might be possible to de-duplicate some code between the original and the parallel version, or even, only use the parallel version by just changing the loop type depending on the availability of rayon.

stephanemagnenat:
Any maintainer to approve the PR?

theotherphil:
Implemented in https://github.com/image-rs/imageproc/pull/527

PR #527, opened 2023-06-06
Adds support for parallel template matching

stephanemagnenat:
See #524 for a discussion of this feature.

There are two commits, the first one adds the new function, and the second one modifies the example to optionally use the new function.

If rayon is disabled, the new feature is also disabled.


theotherphil:
Thanks!

PR #528, opened 2023-06-06
Fix Gaussian blur loosing value by normalizing the finite Gaussian kernel.

stephanemagnenat:
This corrects issue #529 by re-normalizing the Gaussian kernel. This breaks regression tests based on existing images, obviously.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -295,6 +295,8 @@ fn gaussian_kernel_f32(sigma: f32) -> Vec<f32> {
         kernel_data[kernel_radius + i] = value;
         kernel_data[kernel_radius - i] = value;
     }
+    let sum: f32 = kernel_data.iter().sum();
+    kernel_data.iter_mut().for_each(|x| *x /= sum);
     kernel_data
 }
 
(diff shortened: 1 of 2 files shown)


stephanemagnenat:
Of course, if we want compatibility, we could keep the current functions as is and add `_normalized` versions.

theotherphil:
Thanks for the PR.

Please can you move the new tests you've added to regression.rs into the tests section of src/mod/filter.rs, and regenerate the expected outputs for the existing gaussian blur tests in regression.rs by running cargo test with the REGENERATE environment variable set: https://github.com/image-rs/imageproc/blob/4060ac1b28c694b66c4e28fb76d66685c1971647/tests/regression.rs#L8.

theotherphil:
This is a useful fix and I suspect that I ignored the PR for long enough for @stephanemagnenat to lose interest! So I’ll merge as is and then fix the tests in a follow up PR.

stephanemagnenat:
Sorry I am busy with professional deadlines these days, feel free to merge anything that is helpful :-)

PR #534, opened 2023-07-18
Optimize `find_contours_with_threshold`

cospectrum:
Replace expensive allocation `vec![vec![0i32; height]; width]` with more efficient `vec![0i32; height * width]`


cospectrum:
I tried with a couple of images, it was about 40% faster.  And `x + width * y` indexing is faster than `y + height * x`

theotherphil:
Thanks!

Issue #518, opened 2023-04-22
Calculate the area of a contour


joshuajbouw:
Currently there is no way to be able to calculate the area with a contour without doing it manually, which I am struggling with. This will be able then to be used by users to enable the ability to find the minimum and maximum contour in a set of contours.

Calculating the area of a contour is helpful in image detection by finding the largest contour when using the warp perspective method in order to display a wanted part of an image when the background is solid. Unfortunately, this is lacking in this library.

PR #535, opened 2023-07-19
add contour_area

cospectrum:
This will solve [#518](https://github.com/image-rs/imageproc/issues/518).
`oriented_contour_area` will first check the provided points, if there are no points, 1 point or 2 points (line) then the result will be 0.0, otherwise [triangle formula](https://en.wikipedia.org/wiki/Shoelace_formula) is used with the last difference computed first.
`contour_area` will return the absolute value of the `oriented_contour_area`. I don't like the boolean flag for orientation as in OpenCV, because it will be very easy to confuse it

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -68,6 +68,36 @@ where
     res
 }
 
+/// Calculates the area of the contour using the [shoelace formula].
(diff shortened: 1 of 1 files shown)


theotherphil:
Thanks!

PR #538, opened 2023-08-11
Create SECURITY.md

kuzeyardabulut:


jayvdb (commented):
This PR is currently only a template - doesnt have any answers to the questions. 
I dont see this file anywhere under https://github.com/image-rs , so there isnt a suitable example to copy.

IMO this PR can be closed at the moment .  https://github.com/image-rs/image is the 'main' project where this discussion should be done , as it has active maintainers and more users, and has actually had RUSTSEC advisories https://osv.dev/list?ecosystem=&q=image-rs which I dont believe imageproc has had yet.

theotherphil:
Thanks for the contribution, but I agree with @jayvdb’s comments so won’t be merging this.

Issue #540, opened 2023-09-17
`rusttype` dependency is unmaintained, consider migrating to `ab_glyph`


ciuncan:
Below RUSTSEC advisory come up with `cargo audit` for `rusttype` being a transitive dependency of this `imageproc` crate.
https://rustsec.org/advisories/RUSTSEC-2021-0140.html

There in the advisory, it mentions the alternative crate `ab_glyph` which started as fork of `rusttype` and has slight performance increase, too.

PR #541, opened 2023-09-17
Migrate to `ab_glyph`

ciuncan:
Fixes https://github.com/image-rs/imageproc/issues/540

There is a slight difference from previous implementation, that difference is obtained via below command (uses `imagemagick`) and added to truth folder as [`text_delta.png`](https://github.com/ciuncan/imageproc/blob/3570e71d17aacdfeb0855a534a7f8c106dcebb8b/tests/data/truth/text_delta.png) for inspection of this PR.
```bash
compare -compose difference tests/data/truth/text{_rusttype,_ab_glyph,_delta}.png
```

[First commit](https://github.com/image-rs/imageproc/pull/541/commits/198d777c054ebfdcf9cf70bc8fe05bd372e1b038) contains the migration changes, [second one](https://github.com/image-rs/imageproc/pull/541/commits/3570e71d17aacdfeb0855a534a7f8c106dcebb8b) is from `cargo fmt`.


src/drawing/text.rs
```diff
@@ -3,37 +3,50 @@ use crate::drawing::Canvas;
...
-    for g in font.layout(text, scale, point(0.0, v_metrics.ascent)) {
-        if let Some(bb) = g.pixel_bounding_box() {
-            w = max(w, bb.max.x);
-            h = max(h, bb.max.y);
             f(g, bb);
         }
     }
 
-    (w, h)
+    (w as u32, h as u32)
```
fosskers, line 43: Just double-checking that:

- The original `w` and `h` can never be negative?
- The precision loss in converting from floats to ints is acceptable?

ciuncan: Actually I am not 100% sure as it would depend on the coordinates and bounds returned by the `ab_glyph` library, but since these are width and height of the text bounds, I'd expect it to be always non-negative, but just for the peace of mind we can keep them as `i32` instead. As for precision loss, I think it should be the same with previous implementation, but since this is a private function, we could return `f32` and carry the precision just a little more.

src/drawing/text.rs
```diff
@@ -3,37 +3,50 @@ use crate::drawing::Canvas;
...
+                w += kern_width;
+            }
+            last = Some(glyph_id);
+            let bb = g.px_bounds();
+            let g_height = bb.height();
 
-    let (mut w, mut h) = (0, 0);
+            if g_height > h {
+                h = g_height;
+            }
```
ciuncan, line 37: this might be better expressed as:
```suggestion
            h = h.max(g_height);
```

src/drawing/text.rs
```diff
@@ -42,33 +55,30 @@ pub fn text_size(scale: Scale, font: &Font, text: &str) -> (i32, i32) {
...
+pub fn draw_text_mut<C>(
+    canvas: &mut C,
     color: C::Pixel,
-    x: i32,
-    y: i32,
-    scale: Scale,
-    font: &'a Font<'a>,
-    text: &'a str,
+    x: u32,
+    y: u32,
```
ciuncan, line 62: I thought it would make more sense make these coordinates non-negative, but a user of this library might decide to draw string starting from out of bounds of this image continuing into it. We are already doing a bounds check at 78, so it might be better to keep them as `i32`

fosskers: Yeah let's keep them at `i32` and see what happens. That would also be less API breakage.

ciuncan: Agreed, I updated :+1:

fosskers (approved):

fosskers:
While we're at it, want to quickly bump the dependencies again? `cargo upgrade` (from `cargo-edit`) would do it.

ciuncan:
Done :ok_hand:

fosskers (approved):

fosskers:
@theotherphil This should be good to go.

jayvdb:
ping @theotherphil

jayvdb:
@bvssvni , any chance you could review this PR?

bvssvni:
Looks good to me! I am not sure why @theotherphil is not responding.

I opened up https://github.com/image-rs/imageproc/issues/547 just in case this have not been brought up.

theotherphil:
Thanks for the PR and sorry for the slow response.

PR #542, opened 2023-09-28
chore: update dependencies

fosskers:
More dependency updates.

The switch from `rusttype`, which is unmaintained, to `ab_glyph` seems non-trivial.


fosskers:
I hadn't seen this: https://github.com/image-rs/imageproc/pull/541

Issue #478, opened 2021-11-27
PR testing broken


theotherphil:
Travis integration stopped working at some point. Replace with GitHub actions

PR #549, opened 2024-03-10
Add CI/CD

cospectrum:
Will close https://github.com/image-rs/imageproc/issues/478 .
`.github/workflows` derived from https://github.com/jonhoo/rust-ci-conf with some modifications.

Fix clippy, fmt, doc, wasm32 build (getrandom), so that all checks pass. MSRV is `1.63.0`, but for testing the `nightly` version is used

### DIFF
--- a/src/contrast.rs
+++ b/src/contrast.rs
@@ -521,7 +521,8 @@ mod tests {
     fn bench_equalize_histogram_mut(b: &mut Bencher) {
         let mut image = gray_bench_image(500, 500);
         b.iter(|| {
-            black_box(equalize_histogram_mut(&mut image));
+            equalize_histogram_mut(&mut image);
+            black_box(());
         });
     }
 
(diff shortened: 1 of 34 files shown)


src/contrast.rs
```diff
@@ -521,7 +521,8 @@ mod tests {
     fn bench_equalize_histogram_mut(b: &mut Bencher) {
         let mut image = gray_bench_image(500, 500);
         b.iter(|| {
-            black_box(equalize_histogram_mut(&mut image));
+            equalize_histogram_mut(&mut image);
+            black_box(());
```
theotherphil, line 525: This looks pretty suspect when written like this!

theotherphil:
Wow, that’s a very comprehensive set of CI checks! Thanks a lot for this. One day I will need to learn how to use GHA, but that day is not today!

PR #555, opened 2024-03-11
Update MSRV and replace "minimal-versions" check with "direct-minimal-versions" test

cospectrum:
`-Zminimal-versions` + `cargo check` will validate only default features without `dev-dependencies`.
But `-Zminimal-versions` + `cargo +nightly test --all-targets --all-features` is completely broken due to indirect dependencies.
We can try to update with `-Zdirect-minimal-versions` instead and then test.

MSRV updated from `1.63` to `1.65` for `property-testing` feature. Added `--all-features` to the `msrv` workflow (note: additional `--all-targets` flag will not work due to `nightly` tests).

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -98,7 +98,7 @@ jobs:
     # https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
     strategy:
       matrix:
-        msrv: ["1.63.0"] # 2021 edition requires 1.56
+        msrv: ["1.65.0"] # 2021 edition requires 1.56
     name: ubuntu / ${{ matrix.msrv }}
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -6,8 +6,6 @@
 # - hack: check combinations of feature flags
 # - msrv: check that the msrv specified in the crate is correct
 # - semver: check API changes for semver violations.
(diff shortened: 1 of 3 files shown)


cospectrum:
I'm actually not sure if we should update MSRV to `1.65`. But this way we will support one MSRV for `all-features`

CommanderStorm:
> not sure if we should update MSRV to `1.65`

Given that [image-rs has a MSRV of `1.67.1`](https://github.com/image-rs/image/blob/main/CHANGES.md) and that they are likely used together I don't think this should be a roadblock, or is there something I am missing/misunderstanding?

cospectrum:
> > not sure if we should update MSRV to `1.65`
> 
> Given that [image-rs has a MSRV of `1.67.1`](https://github.com/image-rs/image/blob/main/CHANGES.md) and that they are likely used together I don't think this should be a roadblock, or is there something I am missing/misunderstanding?

True, updating to the new `image` = "0.25" requires `1.67.1` rustc. We are currently dependent on "0.24.7", but it's probably a good idea to keep up

.github/workflows/check.yml
```diff
@@ -100,7 +98,7 @@ jobs:
     # https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
     strategy:
       matrix:
-        msrv: ["1.63.0"] # 2021 edition requires 1.56
+        msrv: ["1.65.0"] # 2021 edition requires 1.56
```
theotherphil, line 101: This comment is not very helpful. Better to just delete it?

PR #557, opened 2024-03-15
Re-export `image`

cospectrum:
The problem is described [here](https://www.lurklurk.org/effective-rust/re-export.html).

This gives the user the ability to not depend on `image`, so the code will always compile.
Once this version of `imageproc` is published on `crates.io`, we can add `use imageproc::image` to the beginning of all `examples`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -67,3 +67,5 @@ pub mod template_matching;
 pub mod union_find;
 #[cfg(feature = "display-window")]
 pub mod window;
+
+pub use image;


theotherphil:
Ah, neat - I haven’t come across this before. Thanks.

I’ll publish a new crate version tomorrow.

cospectrum:
> Ah, neat - I haven’t come across this before. Thanks.
> 
> I’ll publish a new crate version tomorrow.

Yes, I already see people trying to use image `0.25` with `imageproc`, but it doesn't compile.

About reexport. I'm wondering if we should add more image features. Because without `"image/default"`, some examples will panic (decoding PNG or other formats) if "imageproc::image" is used.

PR #559, opened 2024-03-15
add `image/default` as default feature

cospectrum:
Continuing our discussion about the re-export of the `image` https://github.com/image-rs/imageproc/pull/557.

`image/default` should allow to run examples using `imageproc::image`, otherwise a panic/error may occur (png/jpeg decode, etc).
It seems to me that users will expect this from our re-export.
If they don't need the default `image` features, then they will do `imageproc = { ..., default-features = false }` which will lead to `image = { ..., default-features = false } ` internally.

The `msrv` is increased, see https://github.com/image-rs/image/issues/2168

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -98,7 +98,7 @@ jobs:
     # https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


theotherphil:
Thanks! Let me know if there’s anything you think needs to go in before the next release, otherwise I’ll just ship what we’ve got around 12 hours from now.

PR #562, opened 2024-03-30
Drop conv dependency

paolobarbolini:
Drops the ancient `conv` dependency, which isn't necessary anymore these days.

Unfortunately this is a breaking change as the `ValueInto` trait appears in many places in the API https://docs.rs/imageproc/0.24.0/imageproc/index.html?search=ValueInto

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ display-window = ["sdl2"]
 [dependencies]
 ab_glyph = "0.2.23"
 approx = "0.5"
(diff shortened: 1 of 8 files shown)


theotherphil:
Thanks! I can no longer remember why I added this dependency in the first place.

cospectrum:
Then let's increase the version to `0.25`? and safety check failed because of libsdl2-2.0-0 installation probably, I'll take a look in free time.

Issue #519, opened 2023-04-28
Panicked while drawing huge ellipse.


eternal-io:
## Very easy to reproduce:

``` rust
#[cfg(test)]
mod tests {
    #[test]
    fn test() {
        let mut img = image::RgbImage::new(1920, 1080);
        imageproc::drawing::draw_filled_ellipse_mut(
            &mut img,
            (960, 540),
            960,
            540,
            image::Rgb([0, 0, 0]),
        );
    }
}
```

## then ...

``` plaintext
thread 'tests::test' panicked at 'attempt to multiply with overflow', D:\.cargo\registry\src\github.com-1ecc6299db9ec823\imageproc-0.23.0\src\drawing\conics.rs:175:50
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library\std\src\panicking.rs:579
   1: core::panicking::panic_fmt
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library\core\src\panicking.rs:64
   2: core::panicking::panic
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library\core\src\panicking.rs:114
   3: imageproc::drawing::conics::draw_ellipse
             at D:\.cargo\registry\src\github.com-1ecc6299db9ec823\imageproc-0.23.0\src\drawing\conics.rs:178
   4: imageproc::drawing::conics::draw_filled_ellipse_mut<image::buffer_::ImageBuffer<image::color::Rgb<u8>,alloc::vec::Vec<u8,alloc::alloc::Global> > >
             at D:\.cargo\registry\src\github.com-1ecc6299db9ec823\imageproc-0.23.0\src\drawing\conics.rs:139
   5: quadim::tests::test
             at .\src\main.rs:483
   6: quadim::tests::test::closure$0
             at .\src\main.rs:481
   7: core::ops::function::FnOnce::call_once<quadim::tests::test::closure_env$0,tuple$<> >
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc\library\core\src\ops\function.rs:250       
   8: core::ops::function::FnOnce::call_once
             at /rustc/84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc/library\core\src\ops\function.rs:250 
```

[...]

PR #564, opened 2024-03-30
Fix panic when drawing a huge ellipse

cospectrum:
Should fix https://github.com/image-rs/imageproc/issues/519

### DIFF
--- a/src/drawing/conics.rs
+++ b/src/drawing/conics.rs
@@ -331,9 +331,6 @@ mod tests {
         fn is_inner_point(&self, (x, y): (i32, i32)) -> bool {
             self.normalized_distance_from_center((x, y)) < 1.0
         }
-        fn is_outer_point(&self, (x, y): (i32, i32)) -> bool {
-            self.normalized_distance_from_center((x, y)) > 1.0
-        }
     }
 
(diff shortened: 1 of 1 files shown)


src/drawing/conics.rs
```diff
@@ -310,8 +310,15 @@ where
 
 #[cfg(test)]
 mod tests {
+    use super::draw_filled_ellipse_mut;
     use image::{GrayImage, Luma};
 
+    #[test]
+    fn draw_huge_ellipse() {
```
theotherphil, line 317: Without an explanation this test with no assertions is pretty unclear.

This could be improved with a comment or by changing the test name, but ideally you could add assertions that some points along the ellipse are drawn - maybe the horizontal and vertical min/max points?

cospectrum: It's just a panic if an old version of a function is used.

theotherphil: I meant that it wouldn’t be clear why this test existed for someone reading it in future.

cospectrum: Ok, I wrote a validation for a filled ellipse

theotherphil:
This is great, thanks!

Issue #545, opened 2023-11-05
BresenhamLineIter crash


erwanvivien:
I found this code to be panicking (because of out-of-bound access)

```rust
let (from, to) = ((240.49998, 0.5), (12.726593, 164.87411));
let image = GrayImage::from_pixel(480, 480, Luma::black());
for _ in BresenhamLinePixelIter::new(&image, from, to) {
    //
}
```

Crashes with 
```
thread 'main' panicked at 'Image index (240, 4294967295) out of bounds (480, 480)'
C:\Users\XXX\.cargo\registry\src\index.crates.io-6f17d22bba15001f\imageproc-0.23.0\src\drawing\line.rs:123:33
```

Has anyone an idea?

erwanvivien:
Added a debug in [`src/drawing/line.rs:66`](https://github.com/image-rs/imageproc/blob/b794265/src/drawing/line.rs#L66):
```
[imageproc\src\drawing\line.rs:67] ret = (
    239,
    0,
)
[imageproc\src\drawing\line.rs:67] ret = (
    240,
    -1,
)
```

And then [`src/drawing/line.rs:123`](https://github.com/image-rs/imageproc/blob/b794265/src/drawing/line.rs#L123) casts it to u32

So changing the positions to ⬇️ fixes it
```rust
let (from, to) = (
    (240.49998f32.floor(), 0.5f32.floor()),
    (12.726593f32.floor(), 164.87411f32.floor()
);
```

I think Brehendam should not take `f32` if it cannot handle it properly (we cast to i32 directly)
Other code like [`draw_antialiased_line_segment`](https://github.com/image-rs/imageproc/blob/b794265/src/drawing/line.rs#L222) already takes `(i32, i32)`

erwanvivien:
Thanks to both of you :)

PR #565, opened 2024-04-08
Fix `BresenhamLineIter` panic

cospectrum:
Should fix https://github.com/image-rs/imageproc/issues/545


theotherphil:
Thanks for the fix.

PR #566, opened 2024-04-09
clamp coverage value in text.rs between 0.0 and 1.0;

asibahi:
Very minimal change. I had copied the code from this crate to use in my own projects, and found out that sometimes `ab_glyph` gives a higher coverage value than `1.0` when there are overlapping glyphs (like in Arabic.) The same change worked in my code. You can find the relevant issue here: https://github.com/alexheretic/ab-glyph/issues/101


theotherphil:
Thanks!

Issue #567, opened 2024-04-10
Template matching with a mask


F8RZD:
open cv provides a mask argument for their template matching function 

mask specifies the pixels that can be ignored during the matching progress 

it's very useful for transparent or white background image template matching

I couldn't find any similar functionality here
 is it planned to be implemented?

any help would be appreciated

cospectrum:
I'll try to implement

cospectrum:
There is no need to guess

<img width="784" alt="template_matching_opencv" src="https://github.com/image-rs/imageproc/assets/96136086/1421081c-29ba-4856-b68d-f2253079644e">

PR #568, opened 2024-04-21
Match template with mask

cospectrum:
2 new functions: `match_template_with_mask` and `match_template_with_mask_parallel`

To close https://github.com/image-rs/imageproc/issues/567


cospectrum:
Tests generated with OpenCV

theotherphil:
Please add a note in the doc comments for the new functions explaining how the mask is used. Otherwise LGTM, thanks!

cospectrum:
> Please add a note in the doc comments for the new functions explaining how the mask is used. Otherwise LGTM, thanks!

Should I write it as a latex formula using [katexit](https://github.com/termoshtt/katexit), for example?

theotherphil:
Oh, that’s cool - I’ve not seen LaTeX used with docs.rs before.

It would make the code harder to read directly for people who don’t know LaTeX, but how can I resist properly rendered formulae in docs!

cospectrum:
@theotherphil 

Without `mask`.

match

With `mask`.

match_masked

Cargo.toml
```diff
@@ -33,6 +33,7 @@ rand_distr = "0.4.3"
 rayon = { version = "1.8.0", optional = true }
 quickcheck = { version = "1.0.3", optional = true }
 sdl2 = { version = "0.36", optional = true, default-features = false, features = ["bundled"] }
+katexit = "0.1.4"
```
ripytide, line 36: Should this not be a `dev-dependency`?

cospectrum: Doesn't work, I tried

theotherphil: I'd rather not add this as a required dependency for all users. I think this can be avoided by following the approach from https://github.com/CEED/libCEED/blob/main/rust/libceed/Cargo.toml

cospectrum: I hope that "katexit" feature will be enabled during the "cargo publish"

theotherphil:
Great, thanks @cospectrum !

PR #569, opened 2024-04-21
Remove unnecessary allocations

cospectrum:
The only public change: `convex_hull`.

### DIFF
--- a/src/filter/median.rs
+++ b/src/filter/median.rs
@@ -252,7 +252,7 @@ impl HistSet {
     fn new(num_channels: u8, expected_count: u32) -> HistSet {
         // Can't use vec![[0u32; 256], num_channels as usize]
         // because arrays of length > 32 aren't cloneable.
-        let mut data = vec![];
+        let mut data = Vec::with_capacity(num_channels as usize);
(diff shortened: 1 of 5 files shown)


theotherphil:
Thanks for the PR. Does replacing a nested loop with Image::from_fn have any impact on the image filtering benchmarks?

cospectrum:
`Image::from_fn` iterates in the same order (row-major, cache friendly), so it should be the same. Plus, you don't have to think about the row-major vs col-major every time.

cospectrum:
> `Image::from_fn` iterates in the same order (row-major, cache friendly), so it should be the same. Plus, you don't have to think about the row-major vs col-major every time.

Hmm, not really. A small regression. It seems that the compiler cannot optimize the mutable lambda

cospectrum:
Or the iterators are to blame.

cospectrum:
A couple of minutes. I'll refactor the lookup tables incrementally

cospectrum:
@theotherphil, Now I think my `bilateral_filter` bench is better than the `master` (both allocs and speed).

theotherphil:
Great, thanks!

PR #570, opened 2024-04-26
Template matching with an alpha channel

daladim:
This MR is adding the `match_template_with_alpha` (and its `match_template_parallel_with_alpha` counterpart).
This make this crate support "masked" templates (or "non rectangular" templates or templates "with an alpha channel", name it as you'd like :D )

A few notes:
* In order not to change the public API, I _added_ new functions, instead of making the same function accept more data types (which would boil down to exposing the `_inner` function).
* I'm not sure I am normalizing the right way (in `sum_squares`) in the case of an image with alpha. I'd be happy to have a review on this part (it's late and my brain is unable to get the meaning of the math behind this normalization, especially as the normalization term depending on the location of the image)

Neat side effects:
* This implementations de-duplicates (a bit) the code between `match_template` and `match_template_parallel`
* I'm making sure the `_parallel` functions are tested in CI
* I'm adding the `gray_alpha_image!` macro

### DIFF
--- a/src/template_matching.rs
+++ b/src/template_matching.rs
@@ -26,6 +26,7 @@ pub enum MatchTemplateMethod {
 }
 
 struct TemplateGrayImage<'a>(&'a GrayImage);
+struct TemplateGrayAlphaImage<'a>(&'a GrayAlphaImage);
 
 trait TemplateImage {
     type InnerBuffer;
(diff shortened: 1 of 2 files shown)


theotherphil:
This and https://github.com/image-rs/imageproc/pull/568 are both adding masked template matching.

cc @daladim and @cospectrum

cospectrum:
> This and #568 are both adding masked template matching.
> 
> cc @daladim and @cospectrum

My PR is compatible with OpenCV, which is a priority for me personally.

theotherphil:
@cospectrum by “compatible with” do you mean being able to produce equal results, or something else?

cospectrum:
> @cospectrum by “compatible with” do you mean being able to produce equal results, or something else?

Equal results, according to their formulas. https://github.com/image-rs/imageproc/issues/567#issuecomment-2053753777

daladim:
I did not see the open issue before implementing. Nice that the formulae for normalisations are posted there (and I realize that I also forgot to square the masked value for the correlation version).
I'll try to compare both MRs tomorrow and see if I can make one with the best of both worlds.

daladim:
@cospectrum 's version is arguably better than mine :)
I'd be in favor of merging his as-is...before a third guy creates an MR with a third implementation!

I may try to create a separate MR to add tests for the `match_template_parallel` (currently, no checks are ensuring its results are consistent with the sequential version).
And I won't merge my `gray_alpha_image!` macro, as no one would use it for now

daladim:
I may as well add a helper function that extracts the template from an image with an alpha channel, to be more convenient when loading a PNG file

daladim:
Before @cospectrum merges his, I've added tests for the parallelized version at https://github.com/cospectrum/imageproc/pull/1

daladim:
This MR could be closed now I've created 
* https://github.com/image-rs/imageproc/pull/583
* https://github.com/cospectrum/imageproc/pull/1
* https://github.com/image-rs/imageproc/pull/584

PR #571, opened 2024-04-27
Add `assert!` for image dims in `bilateral_filter`

cospectrum:
Someone can create an image with `dim > i32::MAX as u32` and `dim as i32` will be negative.


cospectrum:
I checked, there is an `assert!(min <= max)` in `i32::clamp`. But you can still merge.

theotherphil:
Thanks

Issue #314, opened 2019-02-10
Implement L2 Norm for morphology transforms


0b01:
Context: the API for L2 norm is different so it's not implemented as a variant of [`Norm`](https://docs.rs/imageproc/0.17.0/imageproc/distance_transform/enum.Norm.html). I looked at the code it appears conjugating with inverse works for L2 norm.

```rust
fn inverse_mut(image: &mut GrayImage) {
    for i in image.iter_mut() {
        *i = if *i != 0 { 0 } else { 255 };
    }
}

fn erode_l2norm(image: &GrayImage, k: f64) -> GrayImage {
    let mut out = image.clone();
    inverse_mut(&mut out);
    let dist_image = euclidean_squared_distance_transform(&out);
    for (i, o) in dist_image.iter().zip(out.iter_mut()) {
        *o = if *i <= k { 255 } else { 0 };
    }
    inverse_mut(&mut out);
    out
}
```

rickyhan:
Updated version:

```rust
fn erode_l2norm(image: &GrayImage, k: f64) -> GrayImage {
    let mut out = image.clone();
    inverse_mut(&mut out);
    let dist_image = euclidean_squared_distance_transform(&out);
    for (i, o) in dist_image.iter().zip(out.iter_mut()) {
        *o = if *i <= k { 0 } else { 255 };
    }
    out
}
```

Morgane55440:
after looking at the code, it appears obvious to me that the only code that requires change is that of `distance_transform_impl`. that doesn't seem to hard the way i would go around it would be to make three functions (1 for each norm) and simplify `distance_transform_impl` to a match statement to redirect it to the correct implementation. that would also allow the code to use less "if" statements per pixels.
also, i think that, even though it is not too hard to understand after you look at the details, adding a comment explaining why almost all the function is inside a big `unsafe` block would be good, something like : 
```rust
//this code is unsafe because we ignore bounds checking for optimisation reasons
```

Morgane55440:
here is my replacement proposition :

```rust
pub enum Norm {
    /// Defines d((x1, y1), (x2, y2)) to be abs(x1 - x2) + abs(y1 - y2).
    /// Also known as the Manhattan or city block norm.
    L1,
    /// Defines d((x1, y1), (x2, y2)) to be sqrt((x1 - x2)^2 + (y1 - y2)^2).
    /// Also known as the euclidian norm.
    L2,
[...]

PR #572, opened 2024-04-27
Implement L2 Norm for morphology transforms

Morgane55440:
this PR has been made following the issue #314 : **Implement L2 Norm for morphology transforms**

this PR seeks to integrate the L2 Norm for morphology operation, by expanding the [`Norm`](https://docs.rs/imageproc/0.17.0/imageproc/distance_transform/enum.Norm.html) Enum to include L2, and updating all the functions which use this Enum to support the change.
before :
```rust
pub enum Norm {
    /// Defines d((x1, y1), (x2, y2)) to be abs(x1 - x2) + abs(y1 - y2).
    /// Also known as the Manhattan or city block norm.
    L1,
    /// Defines d((x1, y1), (x2, y2)) to be max(abs(x1 - x2), abs(y1 - y2)).
    /// Also known as the chessboard norm.
    LInf,
}
```
after:
```rust
pub enum Norm {
    /// Defines d((x1, y1), (x2, y2)) to be abs(x1 - x2) + abs(y1 - y2).
    /// Also known as the Manhattan or city block norm.
    L1,
    /// Defines d((x1, y1), (x2, y2)) to be sqrt((x1 - x2)^2 + (y1 - y2)^2).
    /// Also known as the euclidian norm.
    /// when working in integer distances, we take the smallest 
    /// greater integer value, also know as 'ceiling'
    /// example : d((0,0),(1,2)) = ceil(sqrt(1+2^2)) = ceil(2.236...) = 3
    L2,
    /// Defines d((x1, y1), (x2, y2)) to be max(abs(x1 - x2), abs(y1 - y2)).
    /// Also known as the chessboard norm.
    LInf,
}
```
on top of adding tests for the new norm, this PR also adds a couple tests for the other norms, in particular when [`erode`](https://docs.rs/imageproc/0.17.0/imageproc/morphology/fn.erode.html) and [`dilate`](https://docs.rs/imageproc/0.17.0/imageproc/morphology/fn.dilate.html) are called with a distance `k` of 0.

### DIFF
--- a/src/distance_transform.rs
+++ b/src/distance_transform.rs
@@ -114,8 +114,8 @@ pub(crate) enum DistanceFrom {
 
 pub(crate) fn distance_transform_impl(image: &mut GrayImage, norm: Norm, from: DistanceFrom) {
     match norm {
-        Norm::LInf => distance_transform_impl_linf(image, from),
-        Norm::L1 => distance_transform_impl_l1(image, from),
+        Norm::LInf => distance_transform_impl_linf_or_l1::<true>(image, from),
+        Norm::L1 => distance_transform_impl_linf_or_l1::<false>(image, from),
         Norm::L2 => {
             match from {
                 DistanceFrom::Foreground => (),
@@ -131,21 +131,6 @@ pub(crate) fn distance_transform_impl(image: &mut GrayImage, norm: Norm, from: D
                 .for_each(|(u, v)| *u = v.sqrt().clamp(0.0, 255.0).ceil() as u8);
         }
     }
-}
-
-#[inline(always)]
-fn distance_transform_impl_linf(image: &mut GrayImage, from: DistanceFrom) {
-    // the boolean generic parameter "IS_LINF" allows
-    // for there to be one compiled fuction per norm,
-    // reducing the number of if statements evaluated at runtime
-    distance_transform_impl_linf_or_l1::<true>(image, from)
-}
-#[inline(always)]
-fn distance_transform_impl_l1(image: &mut GrayImage, from: DistanceFrom) {
-    // the boolean generic parameter "IS_LINF" allows
-    // for there to be one compiled fuction per norm,
-    // reducing the number of if statements evaluated at runtime
-    distance_transform_impl_linf_or_l1::<false>(image, from)
 }
 
 fn distance_transform_impl_linf_or_l1<const IS_LINF: bool>(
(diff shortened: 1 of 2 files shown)


src/distance_transform.rs
```diff
@@ -22,6 +22,12 @@ pub enum Norm {
     /// Defines d((x1, y1), (x2, y2)) to be abs(x1 - x2) + abs(y1 - y2).
     /// Also known as the Manhattan or city block norm.
     L1,
+    /// Defines d((x1, y1), (x2, y2)) to be sqrt((x1 - x2)^2 + (y1 - y2)^2).
+    /// Also known as the euclidian norm.
```
theotherphil, line 26:
```suggestion
    /// Also known as the Euclidean norm.
```

theotherphil:
Please can you run cargo fmt on your PR - formatting is checked in the pre-commit tests.

Morgane55440:
> Please can you run cargo fmt on your PR - formatting is checked in the pre-commit tests.

done - just learned cargo fmt is a thing today

src/distance_transform.rs
```diff
@@ -95,8 +113,53 @@ pub(crate) enum DistanceFrom {
...
+            image
+                .iter_mut()
+                .zip(float_dist.iter())
+                .for_each(|(u, v)| *u = v.sqrt().clamp(0.0, 255.0).ceil() as u8);
+        }
+    }
+}
+
+#[inline(always)]
+fn distance_transform_impl_linf(image: &mut GrayImage, from: DistanceFrom) {
```
theotherphil, line 137: As each of these functions is used only used I think it’s better to just inline the calls to distance_transform_linf_or_l1 into distance_transform_impl and delete the two wrapper functions.

And there’s no need for the comment explaining why we’re using const bool.

src/distance_transform.rs
```diff
@@ -120,7 +183,8 @@ pub(crate) fn distance_transform_impl(image: &mut GrayImage, norm: Norm, from: D
                 if y > 0 {
                     check(image, x, y, x, y - 1);
 
-                    if norm == Norm::LInf {
+                    if IS_LINF {
+                        // this 'if' statement will be compiled away because IS_LINF is a const
```
theotherphil, line 187: This and the matching comment below are unnecessary.

theotherphil:
After inlining the unnecessary wrapper functions this looks good to me, thanks!

Morgane55440:
done !

theotherphil:
Thanks!

PR #573, opened 2024-04-27
Added `Miri` to `CI`

cospectrum:
I've separated all the tests and benchmarks where it makes sense. `benches` are ignored.

UB detected in `nalgebra` (`Svd::try_new`) and `rayon` (if feature is enabled), they are ignored.


src/region_labelling.rs
```diff
@@ -275,7 +275,7 @@ mod tests {
 
     // One huge component with eight-way connectivity, loads of
     // isolated components with four-way connectivity.
-    fn chessboard(width: u32, height: u32) -> GrayImage {
+    pub(super) fn chessboard(width: u32, height: u32) -> GrayImage {
```
cospectrum, line 278: It's used in `super::benches`

src/contrast.rs
```diff
@@ -364,6 +364,7 @@ mod tests {
         }
     }
 
+    #[cfg_attr(miri, ignore)]
```
theotherphil, line 367: Opting out for each benchmark individually feels a bit error prone - could you do the tests vs benches module split everywhere?

cospectrum: Sure, just didn't want to scare you with code moves all over the place

theotherphil:
@cospectrum as there’s been a burst of activity in this repo recently and your change touches a lot of files, let’s merge this after fixing the current merge conflict rather.

src/contrast.rs
cospectrum: Then let's do it in the second PR.

cospectrum:
@theotherphil Conflict fixed.

cospectrum:
Note: we want to ignore `benches` because they are slow, not because they don't pass the `Miri` check.

cospectrum:
> @cospectrum is this good to go or do you have more changes pending for this PR?

You can merge

theotherphil:
Thanks a lot for all the work you’re doing getting a proper testing suite in place.

Issue #576, opened 2024-04-28
support grayscale morphology operators


Morgane55440:
currently the operators in the [morphology](https://docs.rs/imageproc/0.17.0/imageproc/morphology/index.html) module only do binary operations (background = 0, foreground = 255).

but when doing morphological operations on actual operations, it is often necessary to do greyscale operations, using the whole range of values available.

i would like to propose to add greyscale variants of all the currently existing functions, as well as a `Mask` Struct to replace the `Norm` arguments, allowing for user defined masks.

if that is something that seems appropriate to add, i would love to work on the implementation myself.

see [the wikipedia article on grayscale morphology](https://en.wikipedia.org/wiki/Mathematical_morphology#Grayscale_morphology)

theotherphil:
That would be a welcome addition - thanks!

theotherphil:
Implemented in https://github.com/image-rs/imageproc/pull/581

PR #581, opened 2024-04-28
grayscale morphology operators support

Morgane55440:
this PR has been made following the issue #576 : **support grayscale morphology operators**

this PR seeks to add new function to the [`morphology`](https://docs.rs/imageproc/latest/imageproc/morphology/index.html#) module that complement the current ones.

the current function only support [`binary morphology`](https://en.wikipedia.org/wiki/Mathematical_morphology#Binary_morphology), and the new functions will add support for [`grayscale morphology`](https://en.wikipedia.org/wiki/Mathematical_morphology#Grayscale_morphology).

to do so, i have added a new struct `Mask`, that can representations masks of arbitrary shape.

at original time of making this PR, i have 1 commit, that implements the `Mask` struct and all the functions i originally thought of making. on the other hand, i have written very little documentation, and no tests, which is what i'm going to be working on next. i would really appreciate any feedback on the code i've written so far


src/morphology.rs
```diff
@@ -349,6 +349,185 @@ pub fn close_mut(image: &mut GrayImage, norm: Norm, k: u8) {
...
+/// A struct representing a mask used in morphological operations
+///
+/// the mask is represented by a list of the positions of its pixels
+/// relative to its center, with a maximum distance of 255
+/// along each axis.
+/// This means that in the most extreme case, the mask could have
+/// a size of 513 by 513 pixels.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Mask {
+    elements: Vec<(i16, i16)>,
```
theotherphil, line 361: I suspect that for most masks used in practise, we'd get better performance by representing the mask as a grayscale image. But let's get something working with tests first and then worry about performance later :-)

src/morphology.rs
```diff
@@ -349,6 +349,185 @@ pub fn close_mut(image: &mut GrayImage, norm: Norm, k: u8) {
...
+                .collect(),
+        }
+    }
+
+    /// todo!()
+    pub fn disk(radius: u8) -> Self {
+        let mut image = GrayImage::new(2 * radius as u32 + 1, 2 * radius as u32 + 1);
+        image.iter_mut().for_each(|p| *p = 0); // might not unnecessary
+        image.put_pixel(radius as u32, radius as u32, Luma::from([u8::MAX]));
+        dilate_mut(&mut image, Norm::L2, radius);
```
theotherphil, line 390: There's a function for doing this directly: https://github.com/image-rs/imageproc/blob/master/src/drawing/conics.rs#L244

src/morphology.rs
```diff
@@ -349,6 +349,184 @@ pub fn close_mut(image: &mut GrayImage, norm: Norm, k: u8) {
...
+        assert!(
+            (image.height() - center_y as u32) < (u8::MAX as u32),
+            "all pixels of the mask must be at most {} pixels from the center",
+            u8::MAX
+        );
+        Self {
+            elements: (0..image.width() as i16)
+                .flat_map(|x| {
+                    (0..(image.height() as i16))
+                        .filter(move |&y| image.get_pixel(x as u32, y as u32).0[0] != 0)
```
theotherphil, line 419: Nearly all (or maybe all?) of your 'move' keywords look unnecessary as all the types you're using are `copy`.

Morgane55440: they are `copy`, which is why i can multiple `move` in the first place, but the compiler is unable to check whether the closure will outlive the 'x', therefore it needs to `move` a copy of the 'x' inside the closure. because the types are copy, this is practically free. (if we didn't move the 'x', the closure would have to carry a reference to it)

if you know of a way to remove them and make the borrow-checker happy, i'm willing to change things, but simply removing them is not an option.

src/morphology.rs
Morgane55440: thanks !

Morgane55440: i just realized that was a horrible way to do things and have rewritten this function in a much smarter way

src/morphology.rs
theotherphil: Can you give me an example of a compiler error when you remove one of the ‘move’s. I’m not understanding something here.

Morgane55440:
```
 error[E0373]: closure may outlive the current function, but it borrows `x`, which is owned by the current function
   --> src\morphology.rs:396:30
    |
396 |                         .map(|y| (x, y))
    |                              ^^^  - `x` is borrowed here
    |                              |
    |                              may outlive borrowed value `x`
    |
note: closure is returned here
   --> src\morphology.rs:391:21
    |
391 | /                     (-(radius as i16)..=(radius as i16))
392 | |                         .filter(move |&y| {
393 | |                             (x.unsigned_abs() as u32).pow(2) + (y.unsigned_abs() as u32).pow(2)
394 | |                                 <= (radius as u32).pow(2)
395 | |                         })
396 | |                         .map(|y| (x, y))
    | |________________________________________^
help: to force the closure to take ownership of `x` (and any other referenced variables), use the `move` keyword
    |
396 |                         .map(move |y| (x, y))
    |                              ++++
```

theotherphil: Ah yes, sorry - I'm talking nonsense!

This possibly reads slightly more clearly using the `cartesian_product` extension method from itertools:

```
       let range = -(radius as i16)..=(radius as i16);
        let elements: Vec<(i16, i16)> = range
            .clone()
            .cartesian_product(range)
            .filter(|(x, y)| {
                (x.unsigned_abs() as u32).pow(2) + (y.unsigned_abs() as u32).pow(2)
                    <= (radius as u32).pow(2)
            })
            .collect();
```

Pretty subjective though.

Morgane55440: thanks a ton ! if i knew that was an option, i would have used it, seems much clearer to me

Morgane55440: i changed it, but after thinking about it i think it could be that this change slows downs the functions a bit because the use of itertools prevents the compiler from figuring out the output vector size automatically. not really a speed critical place especially compared to the application of the masks, but i will try to test it later

Morgane55440:
i have completed the documentation for all the user-facing functions, just have the testing left

Morgane55440:
i have a question regarding what happens when a mask intersected with the image is empty.
there are 2 case which can cause this :
- Empty Mask
- non-empty mask that does not include the center, on an edge/corner
we can of course disalow empty masks, although i don't think it's necessary, but that wouldn't remove this case.
what would remove this case would be forcing the center to always be part of the mask.

but in my opinion it isn't necessary, and some users might want a mask that doesn't include the center.
so we would need to define what happens when the intersection is empty.
currently `grayscale_erode` sets the pixel to 0, and `grayscale_dilate` to 255.
i could add that to the doc, or change the behavior.

Morgane55440:
different behaviors could be :
- panic : a bit extreme, but that's an extreme case
- return the original pixel value : that would have the benefit of not introducing new value, but that would make empty masks equivalent to a single pixel mask that contains it's center, which might be considered a downside

Morgane55440:
after thinking about it, i realized the answer to "what happens when the intersection of a mask with the image is empty" is obvious.

(u8, min) and (u8, max) are both groups, therefore min({}) and max({}) are well defined, and they are the neutral elements of min and max respectively.

therefore `grayscale_erode` should set such a pixel to `255`, and `grayscale_dilate` to `0`. 
this also might be useful for optimization later on to be used as default values before the mask is computed.

i'm going to fix this and add tests concerning this aspect

src/morphology.rs
```diff
@@ -349,6 +351,631 @@ pub fn close_mut(image: &mut GrayImage, norm: Norm, k: u8) {
...
+            .min()
+            .unwrap_or(u8::MAX)]) // default is u8::MIN because it's the neutral element of max
+    });
+    #[cfg(not(feature = "rayon"))]
+    let result = GrayImage::from_fn(image.width(), image.height(), |x, y| {
+        Luma([mask
+            .apply(image, x, y)
+            .map(|l| l.0[0])
+            .min()
+            .unwrap_or(u8::MAX)]) // default is u8::MIN because it's the neutral element of max
```
theotherphil, line 818:
```suggestion
            .unwrap_or(u8::MAX)]) // default is u8::MAX because it's the neutral element of min
```

src/morphology.rs
```diff
@@ -349,6 +351,631 @@ pub fn close_mut(image: &mut GrayImage, norm: Norm, k: u8) {
...
+
+/// applies a morphologic erosion on the image with the given mask
+///
+/// set each pixel to the minimum of all pixels included in the mask at its position.
+///
+/// See the [`grayscale_erode`](fn.grayscale_erode.html) documentation for examples.
+///
+/// note that unlike `erode_mut`, this performs as many allocations as it's non-mut counterpart
+/// and is thus neither more efficient nor faster
+pub fn grayscale_erode_mut(image: &mut GrayImage, mask: &Mask) {
```
theotherphil, line 831: If the operations can’t be done in place then there’s no reason for the `_mut` variants - they only exist to let users choose to reuse allocations for functions where this is possible.

src/morphology.rs
```diff
@@ -349,6 +351,631 @@ pub fn close_mut(image: &mut GrayImage, norm: Norm, k: u8) {
...
+}
+
+impl Mask {
+    /// creates a mask from a grayscale image
+    ///
+    /// a pixel is part of the mask if and only if it is non-zero
+    ///
+    /// `center_x` and `center_y` define the coordinates of the center of the mask.
+    /// They may take any value, including outside of the bounds of the input image,
+    /// but all pixels of the mask must be at most 255 pixels away from the center.
```
theotherphil, line 373: What’s the reason for the restriction to a distance of 255 pixels?

src/morphology.rs
Morgane55440: that makes sense i'll remove them

src/morphology.rs
Morgane55440: it's slightly arbitrary, but mostly to be consistent with the square(), diamond(), and disk() methods as well as the original morphology functions which all use `u8` , and because that's already way more than enough in my opinion

theotherphil: I agree that 255 should be enough, but if this function would still work for arbitrary i16s I think it’s better to remove the restriction rather than require users to do any arithmetic to work out if their inputs are valid.

Morgane55440: that makes sense, but that's not really an option here. the user is never exposed to the `i16` actually, but what they are exposed to is the u32 type, which is the size of the image.

so there are 3 options here : 
- what i'm doing right now
- limit the size of the image without any arithmetic involved (all mask images must be 255x255 or less)
- don't limit the size of the image : then they can be as big as `u32::max`, and i need to switch the internal type from `i16` to `i64`.

currently, the center coordinates are `u8`, which means `i16`  can represent all distances as long as the image isn't too large, but no matter what type `T` i allow for the center coordinates, i will have to use a greater type internally because it needs to be able to represent `T::MAX - T::MIN` and `T::MIN - T::MAX`, and that can't be done with just `T`

Morgane55440: note on "limit the size of the image without any arithmetic involved" : i could put the maximum at 513x513 or even 1000x1000, as long as the distance between any pair of point along each axis is representable by an ` i16`

theotherphil: A limit on the mask size without reference to center_x and center_y sounds good.

Morgane55440: after thinking about it, i decided to go with 513, because that's the smallest value need to be able to represent all outputs of `square()`, `disk()` and `diamond()`.

if you would prefer an other value, i don't really care too much about it.

Morgane55440:
i sometime struggle between returning a result and panicking. 
Do you feel it would be reasonable to make a `try_from_image` function as well that returns an `Option<Mask>` ?

other than that, i feel like, after it is fully reviewed, it would make sense to push this PR as it is, and make another on the subject of optimisation. what's your opinion ?

cospectrum:
Why do you need a Mask in a public API?

cospectrum:
So It's actually an `Image<LumaA<i16>>` with helper methods.

cospectrum:
If the mask were a LumaA Image, the user would have more freedom and we would have less code I think

Morgane55440:
i understand the idea, but the Mask is not just an `Image<LumaA<i16>>` with helper methods.
 it also requires a center (as the `Mask::new` methods shows).

giving the users a `Mask` allows to bundle the 'image' and the 'center' in one struct

Morgane55440:
i will try to think more on a Maskless implementation though

Morgane55440:
but the reasons i made the struct at the start are also because (1) it felt reasonable to have a distinct struct because mask aren't the same as images, and more importantly (2) that way we could have control on how the mask is laid out in memory, for optimization later on.

the first argument is **very** subjective i will admit, but the second one feel like it makes sense to me.

to be clear, from my (admittedly limited) experience working with mathematical morphology, it should be very rare for users to make masks from images, and almost unthinkable to make masks bigger than 20 by 20. so when the user writes

```rust
let eroded = grayscale_erode(&image, &Mask::diamond(3));
```
it doesn't really matter to them what kind of struct Mask is.

what is affected user-side from the existence or not of the `Mask` struct is twofold :
- the function signature : 
```rust
pub fn grayscale_dilate(image: &GrayImage, mask: &Mask) -> GrayImage
```
feels cleaner to me than
```rust
pub fn grayscale_dilate(image: &GrayImage, mask: &GrayImage) -> GrayImage
```
but that's subjective and not a big deal.
- what happens when users actually want a custom mask : as i mentionned before, that's probably a very rare case. that being said, using an image instead of a `Mask` would do allow you to use images of any size (512 pixels+), but it would prevent you from being able to choose the center yourself, forcing you to pad your image with empty pixels on some sides to move it around. so there are clearly pros and cons.

dev-side, removing the struct would remove the `from_image` and `apply` functions, but the 3 mask makers would still be needed, and we would need to do preprocessing on the "mask" image inputs.

i think that's roughly a good view of the subject. i will admit i'm probably a bit biased toward wanting to keep the struct, but ultimately i'd understand if you feel it's best to remove it.

cospectrum:
3 masks makers can be functions that take `GrayImage` and return an alpha image, for example

cospectrum:
Maybe you can keep `struct` if you think that custom masks are rare. Also `struct` protects some invariants

Morgane55440:
that is what i meant by "3 mask makers would still be needed if we removed the struct" the functions would still be there, just not in an impl block

src/morphology.rs
```diff
@@ -349,6 +351,567 @@ pub fn close_mut(image: &mut GrayImage, norm: Norm, k: u8) {
...
+    ///
+    /// // the center may be out of the image bounds
+    /// let ring_mask_outside = Mask::from_image(&ring_mask_base, 8, 8);
+    ///
+    /// // but the input image itself must be at most 511 pixel wide and 511 pixel wide.
+    /// // the code below would panic :
+    /// // let some_mask = Mask::from_image(&GrayImage::new(600, 600), 2, 2);
+    /// # }
+    /// ```
+    pub fn from_image(image: &GrayImage, center_x: u8, center_y: u8) -> Self {
```
cospectrum, line 418: Are you sure about `u8`?
What about `256`, for example?

Doc: They may take any value, including outside of the bounds of the input image.

Morgane55440: i meant any `u8` value, that's not clear so i'll look into rewording it

src/morphology.rs
```diff
@@ -5,7 +5,9 @@
 use crate::distance_transform::{
     distance_transform_impl, distance_transform_mut, DistanceFrom, Norm,
 };
-use image::GrayImage;
+use image::{GenericImageView, GrayImage, Luma};
+use itertools::Itertools;
+use num::pow::Pow;
```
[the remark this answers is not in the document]
Morgane55440: was added automatically by my code editor will remove it in the next commit.

src/morphology.rs
```diff
@@ -349,6 +351,563 @@ pub fn close_mut(image: &mut GrayImage, norm: Norm, k: u8) {
...
+///     0,   0,  99,   0,   0, 222;
+///     0,   0,  99,   0,   0, 222;
+///     0,   0,   0,   0,   0, 222
+/// );
+///
+/// assert_pixels_eq!(grayscale_dilate(&image, &column_mask), column_dilated);
+/// # }
+/// ```
+pub fn grayscale_dilate(image: &GrayImage, mask: &Mask) -> GrayImage {
+    #[cfg(feature = "rayon")]
```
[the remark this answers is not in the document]
cospectrum: We can do it later actually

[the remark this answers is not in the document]
Morgane55440: i am very interested in any different ideas on this.

i have been thinking about optimization, and i think that ultimately it would be idea to parallelize on image_line/mask_line pairs for cache locality, but that will be tested and changed later.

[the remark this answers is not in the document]
cospectrum: I just want users to know if the function will consume the resources of the rayon thread pool or not. Our template matching for example, we have `match_template` and `match_template_parallel`, not 1 function for both depending on the feature.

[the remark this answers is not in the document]
Morgane55440: doing it in another pr sounds great, if we can have finalize a working foundation and work on it afterward that would be great

Morgane55440:
> Maybe you can keep `struct` if you think that custom masks are rare. Also `struct` protects some invariants

that is def part of my thinking

src/morphology.rs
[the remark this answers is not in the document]
cospectrum: Imagine I have a lot of clients (webserver) or `batch` of work. `inputs.par_iter().map(match_template)` gonna be faster than `inputs.par_iter().map(match_template_parallel)`.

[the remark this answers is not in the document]
cospectrum: So I think you can just delete it.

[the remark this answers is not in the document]
Morgane55440: just did that in my last commit

src/morphology.rs
Morgane55440: i updated it to say the value must be between 0 and 255

Morgane55440:
would it be good to merge as-is ?
 it is functional as far as i can tell, and i've done quite a few tests.
i think it'd be best to work on optimisation/internal mask representation in a new PR

theotherphil:
The docs need a bit of work but that can happen before the next release of this crate.

The behaviour looks correct.

Thanks!

PR #584, opened 2024-04-29
Match template with alpha

daladim:
Very quick-and-dirty proposal for `match_template_with_alpha`, a variant of `match_template_with_mask` that accepts a `GrayImage` (i.e. a template "blended" with its mask) rather than a separate mask.

That's useful because it is a turnkey function that one can call after `image::open`ing a PNG template file.
However, this requires copying the template, then calling `slide_window_at` on separate buffers. I'm not sure how much more efficient it would be to re-implement a `slide_window_at` specifically for this case... Do you think that would be overkill?

**Depends on https://github.com/image-rs/imageproc/pull/568 being merged first**

### DIFF
--- a/src/template_matching.rs
+++ b/src/template_matching.rs
@@ -1,5 +1,5 @@
 //! Functions for performing template matching.
-use crate::definitions::Image;
+use crate::{definitions::Image, drawing::Canvas};
 use image::{GenericImageView, GrayImage, Luma, Primitive};
 
 /// Method used to compute the matching score between a template and an image region.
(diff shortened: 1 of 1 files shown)


daladim:
Discussed in https://github.com/image-rs/imageproc/pull/568

theotherphil:
I don't think it's worth bundling up channel extraction and template matching into a single function in this module.

There are some trivial helpers functions in the `map` module for extracting channels - I'd welcome a PR that added functions there to pull out the relevant channels from GrayAlphaImages. Along the lines of https://github.com/image-rs/imageproc/blob/master/src/map.rs#L553

Issue #585, opened 2024-04-29
`Laplacian` is incorrect


cospectrum:
The actual `Laplacian` can have `negative` values, and for a `GrayImage` the max value can be much `higher` than `255`. But the merged version https://github.com/image-rs/imageproc/pull/516 doesn't care about it. So I want to remove the `detector` and at least fix the `filter`. Also approximation of `Laplacian` is not documented, and `OpenCV` uses `[0, 1, 0, 1, -4, 1, 0, 1, 0]` by default.

thecodergus:
I read your points, reviewed the literature I consulted a year ago (Digital Image Processing, Global Edition - Rafael C. Gonzalez and Richard E. Woods) and I say that it is correct as it is. The default value of OpenCV is the adjacency transform, if you read my previous constructor, the use of [0, 1, 0, 1, -4, 1, 0, 1, 0] was present as an optional situation, if OpenCV uses it as default, it was a decision of the developers, in the literature this is a secondary formula. As for what you mentioned about the real Laplacian being able to have negative values, I didn't understand your point very well, in a GrayImage the resulting value will never be greater than 255 or negative.

cospectrum:
In GrayImage they will not be negative, since in this format the pixels are stored as u8. But it doesn't mean that values are correct.

cospectrum:
If the input values are in `[0, 255]`, then after `Laplacian` (with `[0, 1, 0, 1, -4, 1, 0, 1, 1]` kernel) they can be in `[-1020, 1020]`.

cospectrum:
See `horizontal_sobel`, for example. The input is `GrayImage`, but the output is `Image<Luma<i16>>` to make sure we don't loose the value.

thecodergus:
> See `horizontal_sobel`, for example. The input is `GrayImage`, but the output is `Image<Luma<i16>>` to make sure we don't lose the value.

I understand your point, but for what the algorithm proposes and does is correct, no information will be lost

cospectrum:
Do you think `laplacian_filter` is correct? Because I have `variance of laplacian` in the production, and current implementation from imageproc `master` just not gonna work for me, so I will continue to use my own implementation, which does not lose value.

cospectrum:
https://docs.opencv.org/4.9.0/d5/db5/tutorial_laplace_operator.html

[...]

PR #586, opened 2024-04-29
Fix `Laplacian`

cospectrum:
To close https://github.com/image-rs/imageproc/issues/585


thecodergus:
I definitely don't agree with the changes you made, especially removing the laplacian_edge_detector function

theotherphil:
`laplacian_filter` is consistent with the other filters in this file so let’s stick with that. (This function maybe belongs with the filters in gradients.rs but it’s not a gradient so neither place is obviously better.)

Changing the return type to i16 is definitely correct - I should have spotted that before merging.

Changing the kernel used is more a matter of preference - both the existing kernel and the one in your PR are commonly used approximations to the Laplacian. I’d add an enum to this function letting you choose which to use; we just need to come up with sensible names for the two options!

Don’t just delete the edge detector - the output from the Laplacian filter can be scaled to a u8 value range before thresholding.

cospectrum:
@theotherphil There are no academic names for the `Laplacian` approximation. And I don't think we can come up with good names. 

As for the `detector`, I don't think I can fix it with common scaling :). I think `otsu` threshold is not the right tool for the job. We should simply search for `zero` points after the `Laplacian`. Also approximation of `f(x) = laplacian(gaussian(x))` probably should be better than gaussian followed by laplacian in speed and accuracy.

Maybe will reimplement `detector` in another PR with some `math` arguments behind it?
I just don't like that we have a broken `master`.

And in general, I would prefer to have fewer functions, but so that they are 100% correct, than to have a bunch of functions that only partially work. This is the main reason why I use `imageproc` instead of `opencv`.
Users are not stupid, `imageproc` already provides enough building blocks to implement the `laplacian of gaussian` in 2 lines.

thecodergus:
> `laplacian_filter` is consistent with the other filters in this file so let’s stick with that. (This function maybe belongs with the filters in gradients.rs but it’s not a gradient so neither place is obviously better.)
> 
> Changing the return type to i16 is definitely correct - I should have spotted that before merging.
> 
> Changing the kernel used is more a matter of preference - both the existing kernel and the one in your PR are commonly used approximations to the Laplacian. I’d add an enum to this function letting you choose which to use; we just need to come up with sensible names for the two options!
> 
> Deleting the edge detector isn’t necessary - the output from the Laplacian filter can be scaled to a u8 value range before thresholding. @thecodergus where does the use of Otsu thresholding for this come from? Presumably they use absolute values of intensity rather than linear scaling before applying the threshold?

@theotherphil, I agree that using enums is an excellent idea to allow the user to choose between the two variants of the Laplacian. As for the use of Otsu thresholding in the edge detector, it derives from my empirical experience in research and innovation projects. Its purpose is to highlight the edges more visibly in the final image, and although there is no specific theoretical reason to use the Otsu technique, tests over the years have demonstrated its effectiveness. Regarding the idea of using absolute intensity values instead of linear scaling before threshold application, it is certainly something to be considered



> @theotherphil There are no academic names for the `Laplacian` approximation. And I don't think we can come up with good names.
> 
> As for the `detector`, I don't think I can fix it with scaling :). I think `otsu` threshold is not the right tool for the job. We should simply search for `≈zero` points after the `Laplacian`. Also approximation of `f(x) = laplacian(gaussian(x))` probably should be better than gaussian followed by laplacian in speed and accuracy.
> 
> Perhaps those who need the Laplacian `detector` in the `master` will reimplement it in another PR with some `math` proofs behind it? I just don't like that we have a broken `master`. It makes me very sad.
> 
> And in general, I would prefer to have fewer functions, but so that they are 100% correct, than to have a bunch of functions that only partially work. This is the main reason why I use `imageproc` instead of `opencv`. Users are not stupid, `imageproc` already provides enough building blocks to implement the `laplacian of gaussian` or whatever in 2 lines.

@cospectrum, I am confident that we can find good names for the Laplacian variants. Over the weekend, I will review the literature to explore this idea and also consult my professor, whose primary research area is image processing. Thus, I hope we can solve this problem and keep the master branch functioning correctly. Finally, never underestimate the user's capacity. My goal in creating the edge detection function is to provide a complete and well-structured solution so that the user does not have to resort to auxiliary functions or possess specific academic knowledge to use it successfully.

cospectrum:
I don't think supporting multiple approximations of the Laplacian is worth it, `enum` means something finite and different, but there are an infinite number of approximations, we will not cover all of them anyway.
If the user needs something specific, he can call filter3x3 in 1 line.

But we have much bigger problems now, such as 2 incorrect functions with unsound properties in `master`. And I'm sure there are production codebases that depend on `master`, that's why I am here

ripytide:
I agree with favoring correctness over feature completeness. Removing the `laplacian_edge_detector()` function with it's rather arbitrary implementation for now seems like the best way forward as we can always add it back later after it's implementation details have been justified a bit more thoroughly. `imageproc` should try and stay as unopinionated as possible for the greatest flexibility.

theotherphil:
Ok, let’s fix laplacian_filter for now and worry up getting a correct edge detection algorithm later.

Issue #470, opened 2021-10-05
Add Contrast Normalization


Charles-Schleich:
I am keen to implement a contrast normalization based on this wikipedia article:
https://en.wikipedia.org/wiki/Normalization_(image_processing)

`imageproc::contrast` seems like the right place to put the functions for linear and non-linear normalization.

I will work on this over the next couple of days.

theotherphil:
Implemented in https://github.com/image-rs/imageproc/pull/587

PR #587, opened 2024-05-04
Rebased and Refactored #471 + `range()` function

ripytide:
I'm working through testing/reviewing all the PRs currently.

This PR is a rebase+refactor of #471. #471 seems like a fairly standard and objective image processing function. I've refactored it to use `map()` as well as adding a `range()` function to the `stats.rs` module which is used in `normalize_linear()`.

I also refactored the tests and added benches for the new normalize functions.

Merging this would close #471 and #470.

### DIFF
--- a/src/contrast.rs
+++ b/src/contrast.rs
@@ -10,7 +10,7 @@ use rayon::prelude::*;
 use crate::definitions::{HasBlack, HasWhite};
 use crate::integral_image::{integral_image, sum_image_pixels};
 use crate::map::map_pixels_mut;
-use crate::stats::{cumulative_histogram, histogram};
+use crate::stats::{cumulative_histogram, histogram, minmax};
 
 /// Applies an adaptive threshold to an image.
 ///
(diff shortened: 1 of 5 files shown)


theotherphil:
It looks like you’re missing the change to actually use `range` in `normalise_linear`. 

`Range` is a bit of a confusing name here as users will think of the standard library’s Range types. Maybe `MinMax`?

The existing `stretch_contrast` and the new `normalize_linear` are both special cases of a function that specifies both the desired input and output ranges so I’d like them to be unified rather than adding a similar-but-different parallel implementation.

theotherphil:
Thanks for the work reviewing and updating old PRs :-)

src/contrast.rs
```diff
@@ -266,6 +268,90 @@ pub fn equalize_histogram(image: &GrayImage) -> GrayImage {
...
+        .find_map(|(i, x)| if *x != 0 { Some(i as u8) } else { None })
+        .unwrap();
+
+    //equation from https://en.wikipedia.org/wiki/Normalization_(image_processing)
+    let fraction = (new_max - new_min) as f64 / (max - min) as f64;
+    let f = |_, _, pixel: Luma<u8>| {
+        Luma([(((pixel.0[0] - min) as f64 * fraction) + new_min as f64) as u8])
+    };
+
+    map_pixels_mut(image, f);
```
theotherphil, line 310: These closures are not always optimised away by the compiler. I’d be interested in seeing a benchmark for a raw loop vs this closure based version.

ripytide:
I didn't see `stretch_contrast()` but you're right it's very similar. I've combined both `normalize_linear()` and `stretch_contrast()` into `scale_linear()` which takes both the input_min/max and output_min/max which should allow the user to easily accomplish either goal. 

I've also removed `normalize_nonlinear()` as it used `f64` and was quite complicated and I don't think many users are likely to use it so it is best left up to the users that do want it to implement themselves.

src/contrast.rs
ripytide: I benchmarked using `map_pixels_mut()` with the closure vs a direct `for` loop:
Closure:
```
test contrast::benches::bench_scale_linear     ... bench:      51,119 ns/iter (+/- 308)
test contrast::benches::bench_scale_linear_mut ... bench:      50,364 ns/iter (+/- 192)
```
For Loop:
```
test contrast::benches::bench_scale_linear     ... bench:      51,119 ns/iter (+/- 308)
test contrast::benches::bench_scale_linear_mut ... bench:      50,364 ns/iter (+/- 192)
```

Seems identical to me, but maybe I'm benchmarking it wrong.

src/contrast.rs
```diff
@@ -1,14 +1,16 @@
 //! Functions for manipulating the contrast of images.
 
 use std::cmp::{max, min};
+use std::f64::consts::E;
```
theotherphil, line 4: No longer used

src/stats.rs
```diff
@@ -2,8 +2,52 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct MinMax<T> {
+    /// The minimum value
+    pub min: T,
+    /// The maximum value
+    pub max: T,
+}
+
+/// Returns the minimum and maximum values per channel in an image.
+pub fn minmax<P, T>(image: &Image<P>) -> Vec<MinMax<T>>
```
theotherphil, line 18:
```suggestion
pub fn min_max<P, T>(image: &Image<P>) -> Vec<MinMax<T>>
```

src/contrast.rs
```diff
@@ -266,6 +268,85 @@ pub fn equalize_histogram(image: &GrayImage) -> GrayImage {
...
+
+    let input_min: u16 = input_min.into();
+    let input_max: u16 = input_max.into();
+    let output_min: u16 = output_min.into();
+    let output_max: u16 = output_max.into();
+
+    let input_width = input_max - input_min;
+    let output_width = output_max - output_min;
+
+    let f = |_, _, pixel: Luma<u8>| {
```
theotherphil, line 333: `map_subpixels_mut` would let you write this a bit more concisely - it operates on pixel intensities rather than `(x, y, p)` tuples.

src/contrast.rs
```diff
@@ -266,6 +268,85 @@ pub fn equalize_histogram(image: &GrayImage) -> GrayImage {
...
+pub fn scale_linear_mut(
+    image: &mut GrayImage,
+    input_min: u8,
+    input_max: u8,
+    output_min: u8,
+    output_max: u8,
+) {
+    assert!(
+        input_min <= input_max,
+        "input_min must be smaller or equal to input_max"
```
theotherphil, line 318: The division by `input_width` below would panic if they’re equal, so we need `input_min < input_max`.

src/contrast.rs
```diff
@@ -322,68 +403,14 @@ fn histogram_lut(source_histc: &[u32; 256], target_histc: &[u32; 256]) -> [usize
...
-    }
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::definitions::{HasBlack, HasWhite};
     use crate::utils::gray_bench_image;
     use image::{GrayImage, Luma};
+    use itertools::Itertools;
```
theotherphil, line 412: Not used

src/contrast.rs
```diff
@@ -322,68 +403,14 @@ fn histogram_lut(source_histc: &[u32; 256], target_histc: &[u32; 256]) -> [usize
...
+    use test::{black_box, Bencher};
```
theotherphil, line 413: Not used

src/contrast.rs
```diff
@@ -544,6 +571,15 @@ mod tests {
...
         assert_pixels_eq!(expected, actual);
     }
+
+    #[test]
+    fn test_scale_linear() {
+        let input = gray_image!(1u8, 2, 3, 4, 5, 6, 7, 8, 9, 10, 100, 255);
+
+        let expected = gray_image!(10u8, 10, 10, 11, 11, 12, 12, 13, 13, 13, 52, 120);
+
+        assert_eq!(scale_linear(&input, 1, 255, 10, 120), expected);
```
theotherphil, line 581:
```suggestion
        assert_pixels_eq!(scale_linear(&input, 1, 255, 10, 120), expected);
```

This gives a more useful diff if the actual image does not match the expected image.

theotherphil (commented):
Largely LGTM, just a handful of tweaks needed.

src/stats.rs
ripytide: My thoughts on this is that `minmax` is more consistent with `Itertools::minmax` as well as being the same name as the `MinMax` struct. It doesn't really matter which one to be fair though.

ripytide:
Can I ask why the `unused_imports` lints is allowed? I likely would have caught them before committing had the compiler told me they existed.

src/stats.rs
theotherphil: Let’s go with `min_max`; the Itertools name seems odd here.

theotherphil:
Good question. It was added in https://github.com/image-rs/imageproc/pull/549 and I didn’t spot it at the time. cc @cospectrum

theotherphil:
Thanks!

cospectrum:
> Good question. It was added in #549 and I didn’t spot it at the time. cc @cospectrum

Because there are many places with unused imports, and it will take a lot of time to remove them away

theotherphil:
Seems easy enough to fix - I'll create a PR to remove them.

PR #588, opened 2024-05-05
Add changelog checkbox PR template

ripytide:
I've seen this used in a few other project such as https://github.com/georust/rstar and the template seems a very effective tool to make maintenance easier as users making PRs are automatically reminded to add their changes to the changelog.


theotherphil:
I intend to manually update the changelog for each release anyway, so this would add a bit of work for PR creators but not actually save me any effort.

PR #591, opened 2024-05-05
Exclude some slow tests from miri

theotherphil:


theotherphil:
Still pretty slow, but the last two PRs have reduced Miri run time from 21m 30s to a less painful 5m 40s.

Issue #597, opened 2024-05-06
optimisation of grayscale morphology operators


Morgane55440:
i'd like to make a new PR focused on optimisation for the grayscale morphology operators now that #581 has been merged if that's ok.

theotherphil:
Sounds good!

cospectrum:
What kind of optimizations?

cospectrum:
> i'd like to make a new PR focused on optimisation for the grayscale morphology operators now that #581 has been merged if that's ok.

<img width="729" alt="cache_miss" src="https://github.com/image-rs/imageproc/assets/96136086/1bbafd03-d13e-4364-b1d6-3a8a0814f489">

I think you have a lot of cache misses here because you are accessing image pixels in col-major order.

cospectrum:
So you can safely change the order of `elements: Vec<(i16, i16)>` to `row-major`. And I would change `(i16, i16)` to `Point<i16>` for readability.

theotherphil:
@Morgane55440 is this issue complete now or are you still considering the other optimisation ideas you mentioned above?

Morgane55440:
i tink i still would like to try the bit_mask technique. 
honestly i don't expect it to actually be great because the SIMD operation that is likely being used atm(prob something like [_mm512_min_epi8](https://doc.rust-lang.org/beta/core/arch/x86/fn._mm512_min_epi8.html#)) which does byte-by-byte min, is from what i understand faster than computing the minimum byte from a group of bytes, which was my original idea, but there still is a good chance it might actually be faster.
i'll try to do some tests on my machine, if they're inconclusive i'll close

cospectrum:
Simd is architecture dependent, currently imageproc is targeting cross platform CPU I think

theotherphil:
We can remain cross-platform using runtime detection of SIMD support. I've no idea how practical this will end up being, but there's potentially a lot of speedup available for some functions.

Morgane55440:
> Simd is architecture dependent, currently imageproc is targeting cross platform CPU I think

[...]

PR #598, opened 2024-05-08
optimisation of grayscale morphology operators

Morgane55440:
This PR has been made following the issue #597  : optimization of grayscale morphology operators

it aims to work on the internal implementation of the `Mask` and associated functions to optimize the execution time, as well as add parallel versions of some public functions using rayon.

in the first commit that i have made, i have added bench for the `Mask` creation methods and have reworked the implementation of the `from_image` methods, which allowed a 4x speed increase (200 ms to 50ms) for a large image (200 by 200 pixels). i would expect the improvement to be less significant for smaller images, although i have not tested it.

it was brought to my attention that i was accessing image pixels in col-major order. That was due to a misunderstanding of the image layout from my part. In light of this, i have reworked the implementation of the `Mask` to facilitate and incentivize access in row-major order by storing the positions row by row.

i have also been advised to change (i16, i16) to Point<i16> for readability, which i will do in a further commit.

Finally, through my testing, i tried to use `unsafe_get_pixel` to optimize my `get_pixel` calls. While it was very efficient, i found that accessing the buffer directly and using the `chunks` method was faster by around 17% (~65  to ~50microsecond ). i tested it several times, and the slowest `chunks` time was faster than the fastest `unsafe_get_pixel` by 10 ms. 

i believe that it might be interesting to look at for other uses of `unsafe_get_pixel` which traverse the image line by line, as it potentially could be the case that directly traversing a `[u8]` might allow the compiler to do better optimizations than when these accesses are obfuscated through `unsafe_get_pixel`

### DIFF
--- a/src/morphology.rs
+++ b/src/morphology.rs
@@ -627,7 +627,7 @@ impl Mask {
     }
 
     /// all the slices are garanteed to be non empty
-    fn lines<'a>(&'a self) -> impl Iterator<Item = &'a [Point<i16>]> {
+    fn lines(&self) -> impl Iterator<Item = &[Point<i16>]> {
         self.elements.chunk_by(|p1, p2| p1.y == p2.y)
     }
 }
@@ -733,7 +733,7 @@ pub fn grayscale_dilate(image: &GrayImage, mask: &Mask) -> GrayImage {
                 .unwrap_or(0);
 
             for i in 0..image.width() {
-                if l_bound > 0 && i64::from(line[l_bound - 1].x) >= i64::from(i){
+                if l_bound > 0 && i64::from(line[l_bound - 1].x) >= i64::from(i) {
                     l_bound -= 1
                 };
                 if r_bound > 0
--- a/src/morphology.rs
+++ b/src/morphology.rs
@@ -370,7 +370,6 @@ pub struct Mask {
 }
 
 impl Mask {
-
     /// creates a mask from a grayscale image
     ///
     /// a pixel is part of the mask if and only if it is non-zero
--- a/src/morphology.rs
+++ b/src/morphology.rs
@@ -466,13 +466,12 @@ impl Mask {
     /// ```
     pub fn square(radius: u8) -> Self {
         let range = -(radius as i16)..=(radius as i16);
-        Self {
-            elements: range
-                .clone()
-                .cartesian_product(range)
-                .map(|(y, x)| Point::new(x, y))
-                .collect(),
-        }
+        let elements = range
+            .clone()
+            .cartesian_product(range)
+            .map(|(y, x)| Point::new(x, y))
+            .collect();
+        Self { elements }
     }
 
     /// creates a diamond-shaped mask
@@ -515,14 +514,12 @@ impl Mask {
     /// # }
     /// ```
     pub fn diamond(radius: u8) -> Self {
-        Self {
-            elements: (-(radius as i16)..=(radius as i16))
-                .flat_map(|y| {
-                    ((y.abs() - radius as i16)..=(radius as i16 - y.abs()))
-                        .map(move |x| Point::new(x, y))
-                })
-                .collect(),
-        }
+        let mut elements = Vec::new();
+        let mut iter = (-(radius as i16)..=(radius as i16)).flat_map(|y| {
+            ((y.abs() - radius as i16)..=(radius as i16 - y.abs())).map(move |x| Point::new(x, y))
+        });
+        elements.resize_with(1 + 2 * (radius as usize) * (radius as usize + 1), || iter.next().expect("the size of iter is exactly equal to 1 + 2 * radius * (radius + 1) (i, Morgane55440, did the math)"));
+        Self { elements }
     }
 
     /// creates a disk-shaped mask
@@ -582,17 +579,18 @@ impl Mask {
     /// ```
     pub fn disk(radius: u8) -> Self {
         let range = -(radius as i16)..=(radius as i16);
-        Self {
-            elements: range
-                .clone()
-                .cartesian_product(range)
-                .filter(|(y, x)| {
-                    (x.unsigned_abs() as u32).pow(2) + (y.unsigned_abs() as u32).pow(2)
-                        <= (radius as u32).pow(2)
-                })
-                .map(|(y, x)| Point::new(x, y))
-                .collect(),
-        }
+        let radius_squared = (radius as u32) * (radius as u32);
+        let elements = range
+            .clone()
+            .cartesian_product(range)
+            .filter(|(y, x)| {
+                (x.unsigned_abs() as u32) * (x.unsigned_abs() as u32)
+                    + (y.unsigned_abs() as u32) * (y.unsigned_abs() as u32)
+                    <= radius_squared
+            })
+            .map(|(y, x)| Point::new(x, y))
+            .collect();
+        Self { elements }
     }
 
     fn apply<'a, 'b: 'a, 'c: 'a>(
--- a/src/morphology.rs
+++ b/src/morphology.rs
@@ -630,6 +630,36 @@ impl Mask {
     }
 }
 
+fn mask_reduce<F: Fn(u8, u8) -> u8>(
+    image: &GrayImage,
+    mask: &Mask,
+    neutral: u8,
+    operator: F,
+) -> GrayImage {
+    let mut result: image::ImageBuffer<Luma<u8>, Vec<u8>> =
+        GrayImage::from_pixel(image.width(), image.height(), Luma([neutral]));
+    for (y, line_group) in mask.lines().into_iter() {
+        let y = i64::from(y);
+        let x_line = line_group.map(|p| p.x).collect::<Vec<_>>();
+        let input_rows = image
+            .chunks(image.width() as usize)
+            .skip(y.try_into().unwrap_or(0));
+        let output_rows = result
+            .chunks_mut(image.width() as usize)
+            .skip((-y).try_into().unwrap_or(0));
+        for (input_row, output_row) in input_rows.zip(output_rows) {
+            for x in x_line.iter().copied() {
+                for (input, output) in (input_row.iter().skip(x.try_into().unwrap_or(0)))
+                    .zip(output_row.iter_mut().skip((-x).try_into().unwrap_or(0)))
+                {
+                    *output = operator(*input, *output);
+                }
+            }
+        }
+    }
+    result
+}
+
 /// computes the morphologic dilation of the input image with the given mask
 ///
 /// for each input pixel, the output pixel will be the maximum of all pixels included
@@ -714,28 +744,7 @@ impl Mask {
 /// ```
 pub fn grayscale_dilate(image: &GrayImage, mask: &Mask) -> GrayImage {
     // default is u8::MIN because it's the neutral element of max
-    let mut result: image::ImageBuffer<Luma<u8>, Vec<u8>> =
-        GrayImage::from_pixel(image.width(), image.height(), Luma([u8::MIN]));
-    for (y, line_group) in mask.lines().into_iter() {
-        let y = i64::from(y);
-        let x_line = line_group.map(|p| p.x).collect::<Vec<_>>();
-        let input_rows = image
-            .chunks(image.width() as usize)
-            .skip(y.try_into().unwrap_or(0));
-        let output_rows = result
-            .chunks_mut(image.width() as usize)
-            .skip((-y).try_into().unwrap_or(0));
-        for (input_row, output_row) in input_rows.zip(output_rows) {
-            for x in x_line.iter().copied() {
-                for (input, output) in (input_row.iter().skip(x.try_into().unwrap_or(0)))
-                    .zip(output_row.iter_mut().skip((-x).try_into().unwrap_or(0)))
-                {
-                    *output = (*output).max(*input);
-                }
-            }
-        }
-    }
-    result
+    mask_reduce(image, mask, u8::MIN, u8::max)
 }
 
 /// computes the morphologic erosion of the input image with the given mask
@@ -822,28 +831,7 @@ pub fn grayscale_dilate(image: &GrayImage, mask: &Mask) -> GrayImage {
 /// ```
 pub fn grayscale_erode(image: &GrayImage, mask: &Mask) -> GrayImage {
     // default is u8::MAX because it's the neutral element of min
-    let mut result: image::ImageBuffer<Luma<u8>, Vec<u8>> =
-        GrayImage::from_pixel(image.width(), image.height(), Luma([u8::MAX]));
-    for (y, line_group) in mask.lines().into_iter() {
-        let y = i64::from(y);
-        let x_line = line_group.map(|p| p.x).collect::<Vec<_>>();
-        let input_rows = image
-            .chunks(image.width() as usize)
-            .skip(y.try_into().unwrap_or(0));
-        let output_rows = result
-            .chunks_mut(image.width() as usize)
-            .skip((-y).try_into().unwrap_or(0));
-        for (input_row, output_row) in input_rows.zip(output_rows) {
-            for x in x_line.iter().copied() {
-                for (input, output) in (input_row.iter().skip(x.try_into().unwrap_or(0)))
-                    .zip(output_row.iter_mut().skip((-x).try_into().unwrap_or(0)))
-                {
-                    *output = (*output).min(*input);
-                }
-            }
-        }
-    }
-    result
+    mask_reduce(image, mask, u8::MAX, u8::min)
 }
 
 /// Grayscale erosion followed by grayscale dilation.
--- a/src/morphology.rs
+++ b/src/morphology.rs
@@ -845,7 +845,7 @@ pub fn grayscale_erode(image: &GrayImage, mask: &Mask) -> GrayImage {
 /// use imageproc::morphology::{Mask, grayscale_open};
 ///
 /// // Isolated regions of foreground pixels are removed,
-/// // while isolated zones of background pixels are maintaned
+/// // while isolated zones of background pixels are maintained
 /// let image = gray_image!(
 ///   100,  99,  99,  99, 222,  99;
 ///    99,  99,  99, 222, 222, 222;
@@ -855,7 +855,7 @@ pub fn grayscale_erode(image: &GrayImage, mask: &Mask) -> GrayImage {
 /// );
 ///
 /// // Isolated regions of foreground pixels are removed,
-/// // while isolated zones of background are maintaned
+/// // while isolated zones of background are maintained
 /// let image_opened = gray_image!(
 ///    99,  99,  99,  99,  99,  99;
 ///    99,  99,  99,  99,  99,  99;
@@ -910,7 +910,7 @@ pub fn grayscale_open(image: &GrayImage, mask: &Mask) -> GrayImage {
 /// );
 ///
 /// // Isolated regions of background pixels are removed,
-/// // while isolated zones of foreground pixels are maintaned
+/// // while isolated zones of foreground pixels are maintained
 /// let image_closed = gray_image!(
 ///    99,  99,  99, 222, 222, 222;
 ///    99,  99,  99, 222, 222, 222;
--- a/src/morphology.rs
+++ b/src/morphology.rs
@@ -361,7 +361,7 @@ pub fn close_mut(image: &mut GrayImage, norm: Norm, k: u8) {
 /// The mask can have any size between 0 by 0 to 511 by 511 pixels
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Mask {
-    /// for any optimisation/arithmetic purposes, it is garanteed that :
+    /// for any optimisation/arithmetic purposes, it is guaranteed that:
     ///
     /// - all the integer values will be strictly between -512 and 512
     /// - all tuples will be sorted in reverse lexicographic order, line by line ((-1,-1),(0,-1),(1,-1),(-1,0),(0,0),...)
--- a/src/morphology.rs
+++ b/src/morphology.rs
@@ -2,12 +2,14 @@
 //!
 //! [morphological operators]: https://homepages.inf.ed.ac.uk/rbf/HIPR2/morops.htm
 
+use std::slice::Iter;
+
 use crate::{
     distance_transform::{distance_transform_impl, distance_transform_mut, DistanceFrom, Norm},
     point::Point,
 };
 use image::{GrayImage, Luma};
-use itertools::Itertools;
+use itertools::{GroupBy, Itertools};
 
 /// Sets all pixels within distance `k` of a foreground pixel to white.
 ///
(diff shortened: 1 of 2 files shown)


Morgane55440:
next i'm changing the `(i16, i16)` to `Point`, benching `grayscale_erode` and `grayscale_dilate`, and testing different implementations

cospectrum:
What about this? Similar performance
```rs
let elements = image
            .enumerate_pixels()
            .filter(|(_, _, &p)| p[0] != 0)
            .map(|(x, y, _)| (x as i16 - center_x as i16, y as i16 - center_y as i16))
            .collect();
Self { elements }
```

Morgane55440:
> ```rust
> let elements = image
>             .enumerate_pixels()
>             .filter(|(_, _, &p)| p[0] != 0)
>             .map(|(x, y, _)| (x as i16 - center_x as i16, y as i16 - center_y as i16))
>             .collect();
> Self { elements }
> ```

thanks a ton ! it is just as fast, if not slightly faster (seems maybe 5-10 ms faster on average, but there was overlap)

Morgane55440:
i am not super knowledgeable on all the ìmage` methods, but i'll look into them more

Morgane55440:
i think i probably should look into the other constructors

```
running 4 tests
test morphology::benches::bench_grayscale_diamond_mask                                ... bench:     185,236 ns/iter (+/- 19,262)
test morphology::benches::bench_grayscale_disk_mask                                   ... bench:     112,540 ns/iter (+/- 36,400)
test morphology::benches::bench_grayscale_mask_from_image                             ... bench:      47,545 ns/iter (+/- 1,532)
test morphology::benches::bench_grayscale_square_mask                                 ... bench:      43,016 ns/iter (+/- 1,629)
```
(they all make masks of the same size)

cospectrum:
The new release is scheduled for May 14th, if anything. I think you will make it in time

src/morphology.rs
```diff
@@ -511,13 +514,12 @@ impl Mask {
...
-                .flat_map(|x| {
-                    ((x.abs() - radius as i16)..=(radius as i16 - x.abs())).map(move |y| (x, y))
-                })
-                .collect(),
-        }
+        let mut elements = Vec::new();
+        let mut iter = (-(radius as i16)..=(radius as i16)).flat_map(|y| {
+            ((y.abs() - radius as i16)..=(radius as i16 - y.abs())).map(move |x| Point::new(x, y))
+        });
+        elements.resize_with(1 + 2 * (radius as usize) * (radius as usize + 1), || iter.next().expect("the size of iter is exactly equal to 1 + 2 * radius * (radius + 1) (i, Morgane55440, did the math)"));
```
cospectrum, line 521: Don't panic. You can try `Vec::with_capacity` and then for loop with `push`.

Morgane55440: i have done something similar, but with `extend` instead of `push`

Morgane55440:
```
running 4 tests
test morphology::benches::bench_grayscale_diamond_mask                                ... bench:      17,375 ns/iter (+/- 1,134)
test morphology::benches::bench_grayscale_disk_mask                                   ... bench:      33,289 ns/iter (+/- 3,166)
test morphology::benches::bench_grayscale_mask_from_image                             ... bench:      50,766 ns/iter (+/- 2,226)
test morphology::benches::bench_grayscale_square_mask                                 ... bench:      40,882 ns/iter (+/- 2,205)
```
pretty happy with this. going to sleep. Tomorrow i'll work on making the disk mask code clearer, and then i'll move on to `erode` and `dilate`

Morgane55440:
@cospectrum if you don't mind i would really appreciate your take on my last commit, i kinda struggled with making `disk` readable

Morgane55440:
initial stae of benches : 
```
test morphology::benches::bench_grayscale_op_dilate_big_image_diamond                 ... bench:  86,049,620 ns/iter (+/- 13,300,790)
test morphology::benches::bench_grayscale_op_dilate_big_image_point                   ... bench:   2,315,520 ns/iter (+/- 188,327)
test morphology::benches::bench_grayscale_op_dilate_medium_image_diamond              ... bench:   3,373,130 ns/iter (+/- 193,688)
test morphology::benches::bench_grayscale_op_dilate_medium_image_large_square         ... bench: 131,351,560 ns/iter (+/- 8,071,737)
test morphology::benches::bench_grayscale_op_dilate_medium_image_point                ... bench:      99,243 ns/iter (+/- 13,569)
test morphology::benches::bench_grayscale_op_dilate_small_image_diamond               ... bench:     208,150 ns/iter (+/- 14,842)
test morphology::benches::bench_grayscale_op_dilate_small_image_large_square          ... bench:   7,243,630 ns/iter (+/- 1,650,297)
test morphology::benches::bench_grayscale_op_dilate_small_image_point                 ... bench:       5,766 ns/iter (+/- 178)
test morphology::benches::bench_grayscale_op_erode_big_image_diamond                  ... bench:  88,131,170 ns/iter (+/- 9,589,572)
test morphology::benches::bench_grayscale_op_erode_big_image_point                    ... bench:   2,311,410 ns/iter (+/- 726,459)
test morphology::benches::bench_grayscale_op_erode_medium_image_diamond               ... bench:   3,503,005 ns/iter (+/- 314,480)
test morphology::benches::bench_grayscale_op_erode_medium_image_large_square          ... bench: 130,097,690 ns/iter (+/- 6,730,746)
test morphology::benches::bench_grayscale_op_erode_medium_image_point                 ... bench:      82,883 ns/iter (+/- 5,617)
test morphology::benches::bench_grayscale_op_erode_small_image_diamond                ... bench:     212,093 ns/iter (+/- 40,138)
test morphology::benches::bench_grayscale_op_erode_small_image_large_square           ... bench:   7,109,105 ns/iter (+/- 483,316)
test morphology::benches::bench_grayscale_op_erode_small_image_point                  ... bench:       5,180 ns/iter (+/- 449)
```

cospectrum:
> @cospectrum if you don't mind i would really appreciate your take on my last commit, i kinda struggled with making `disk` readable

The `master` version is much more beautiful.

Morgane55440:
> > @cospectrum if you don't mind i would really appreciate your take on my last commit, i kinda struggled with making `disk` readable
> 
> The `master` version is much more beautiful.

it is, but it's 3-4 times slower

Morgane55440:
what i did there was, instead of just checking if each pixel is inside or outside the disk, i computed the edges of the disk, and then filled it in.

it is messier, but much faster. i would like to maybe make it less messy without necessarily sacrificing speed

src/morphology.rs
```diff
@@ -576,17 +577,39 @@ impl Mask {
     /// # }
     /// ```
     pub fn disk(radius: u8) -> Self {
-        let range = -(radius as i16)..=(radius as i16);
-        Self {
-            elements: range
+        let radius_squared = (radius as u32) * (radius as u32);
```
ripytide, line 580: Is there a reason why `clippy::cast_lossless` is an allowed lint? It's quite confusing when reading a lot of casts from `u8` to bigger types like `u16` and `u32` when such a cast should be lossless. It makes it harder to spot lossy casts when checking correctness.

Morgane55440: i know very little about clippy, i have no idea. i just run clippy and do what it tells me because i has always made sense so far. if i should change anything please tell me. in this specific function, i cast he `i16`s as `u32`s to prevent potential overflow from the squaring

ripytide: Apologies my question was more about this projects clippy settings and not your PR (it just happens to be the first PR I've noticed it on), if you're interested though you can read about the lint here: https://rust-lang.github.io/rust-clippy/master/index.html#/cast_loss. In this scenario I would recommend using `u32::from(radius)` or using `x.into()` methods over explicit casts. Ideally, clippy would enforce this for us however which was my wider point although that's not related to this particular PR.

Morgane55440: thanks, that makes a lot of sense, i've tried to use as many `from`s and `into()`s as possible. 

my next commit will take a little while because i've been working on he optimization of `erode` and `dilate` too.

Morgane55440:
the [new commit](https://github.com/image-rs/imageproc/pull/598/commits/62eb25de885a898df0e329d713a589b1608fd155) is faster by about 20-30% for big images, but it is ***utterly unreadable***, i will rework it to make it even faster and significantly more readable and faster

also, i did not know about the minimum supported version, will rewrite the offending code

src/morphology.rs
cospectrum: > Is there a reason why `clippy::cast_lossless` is an allowed lint? It's quite confusing when reading a lot of casts from `u8` to bigger types like `u16` and `u32` when such a cast should be lossless. It makes it harder to spot lossy casts when checking correctness.

I think the reasons are again the same as in the case of `unused imports`. Someone wrote code a long time ago that violates the rules, and it will take some time to rewrite it. Personally, I would remove all `allows` (in lib.rs), and where the rules still need to be violated, let it be localized to a function or module.

ripytide: If that is the case, I 100% agree with your approach. I can draft a PR to make that change but it would be quite a lot of code churn it might cause merge conflicts with some in-flight PRs. Perhaps after the next release if the PRs die down a bit we can give it a shot then.

cospectrum:
Write simple reference implementations and test against them

src/morphology.rs
```diff
@@ -688,13 +721,37 @@ impl Mask {
...
 /// ```
 pub fn grayscale_dilate(image: &GrayImage, mask: &Mask) -> GrayImage {
-    let result = GrayImage::from_fn(image.width(), image.height(), |x, y| {
-        Luma([mask
-            .apply(image, x, y)
-            .map(|l| l.0[0])
-            .max()
-            .unwrap_or(u8::MIN)]) // default is u8::MIN because it's the neutral element of max
-    });
+    let mut result = GrayImage::from_fn(image.width(), image.height(), |_, _| Luma([u8::MIN]));
```
cospectrum, line 724: `GrayImage::from_pixel(width, height, Luma([0]))`

src/morphology.rs
```diff
@@ -688,13 +721,37 @@ impl Mask {
...
+    for (y_16, x_line) in mask.lines() {
```
cospectrum, line 725: I don't like postfixes, just shadow the variable

src/morphology.rs
```diff
@@ -688,13 +721,37 @@ impl Mask {
...
+        for (input_row, output_row) in input_rows.zip(output_rows) {
+            let mut l_bound = x_line.iter().position(|&x| x >= 0).unwrap_or(x_line.len());
+            let mut r_bound = x_line
+                .iter()
+                .rposition(|&x| input_row.len() as i64 > x.into())
+                .map(|x| x + 1)
+                .unwrap_or(0);
+
+            for (i, p) in output_row.iter_mut().enumerate() {
+                let i_i64 = i as i64; // works because image size fits into u32
```
cospectrum, line 742: rm postfix

src/morphology.rs
```diff
@@ -597,12 +619,23 @@ impl Mask {
...
     }
+
+    /// all the slices are garanteed to be non empty
+    fn lines(&self) -> impl Iterator<Item = (i16, Vec<i16>)> + '_ {
+        self.elements
+            .iter()
+            .group_by(|p| p.y)
+            .into_iter()
+            .map(|(y, g)| (y, g.map(|p| p.x).collect()))
+            .collect::<Vec<_>>()
```
cospectrum, line 636: Why collect? Just return after `map`

src/morphology.rs
```diff
@@ -427,13 +428,12 @@ impl Mask {
...
-                .filter(|(x, y)| image.get_pixel(*x as u32, *y as u32).0[0] != 0)
-                .map(|(x, y)| (x - center_x as i16, y - center_y as i16))
-                .collect(),
-        }
+        let elements = image
+            .enumerate_pixels()
+            .filter(|(_, _, &p)| p[0] != 0)
+            .map(|(x, y, _)| Point::new(x as i16 - center_x as i16, y as i16 - center_y as i16))
+            .collect();
+        Self { elements }
```
cospectrum, line 436: Add private `new` and call it instead of `Self { elements }` everywhere

```
fn new(elements: Vec<Point<i16>>) -> Self {
    assert!(elements.len() <= (511 * 511) as usize);
    Self { elements } 
}
```

src/morphology.rs
```diff
@@ -511,13 +514,11 @@ impl Mask {
...
     pub fn diamond(radius: u8) -> Self {
-        Self {
-            elements: (-(radius as i16)..=(radius as i16))
-                .flat_map(|x| {
-                    ((x.abs() - radius as i16)..=(radius as i16 - x.abs())).map(move |y| (x, y))
-                })
-                .collect(),
-        }
+        let mut elements = Vec::with_capacity(1 + 2 * (radius as usize) * (radius as usize + 1));
+        elements.extend((-(radius as i16)..=(radius as i16)).flat_map(|y| {
```
cospectrum, line 518: Add `let radius = i16::from(radius);` above

src/morphology.rs
```diff
@@ -576,17 +577,38 @@ impl Mask {
...
     /// ```
     pub fn disk(radius: u8) -> Self {
-        let range = -(radius as i16)..=(radius as i16);
-        Self {
-            elements: range
+        let radius_squared = u32::from(radius).pow(2);
+        let half_widths_per_height = std::iter::successors(
+            Some((-i16::from(radius), 0u8)),
+            |(last_height, last_half_width)| {
+                if *last_height == i16::from(radius) {
```
cospectrum, line 584: You can add `let radius = i16::from(radius)` before loop and reuse it

cospectrum (commented):
.

theotherphil:
The `chunks` finding is interesting. This crate originally aimed to work on `GenericImage`s but I gave up on that when the performance overhead was too high. Now that we know we’re working with image buffers we can probably find a few places where existing functions can be sped up.

As this PR doesn’t change any APIs and comes with a significant speedup I’d like to get his merged for the next release - let me know when you’re ready for this to be merged. We can always simplify or clarify implementations later if necessary.

cospectrum:
If it works, then I can simplify/refactor later myself, after the merge

src/morphology.rs
Morgane55440: thanks fixed in next commit

src/morphology.rs
Morgane55440: done in next commit

src/morphology.rs
Morgane55440: not sure i fully understand, but basically it's a lifetime issue with GroupBy. Groupby is an iterable, not an iterator, thus issues with ownership

error_groupby

src/morphology.rs
Morgane55440: done !

src/morphology.rs
Morgane55440: done, as well as in other functions where the problem appeared

src/morphology.rs
Morgane55440: done in next commit

Morgane55440:
@theotherphil @cospectrum sorry for the wait was busy with life stuff. i finally got a decent optimization in, from 2x to almost 40x depending on the image and mask size/shape (the bigger the better the optimization)

i'm certain significant further improvements can be made, but i'm ok with shipping things as they are, except for the `lines` method which is a mess. i'm working on it. 

also the `apply` method is now unused. i think it's possible it might be useful later although it seems unlikely because it's just a very slow way to do things. should i remove it ?

theotherphil:
Brilliant, thanks. Yes, remove `apply` if it’s not used.

Morgane55440:
i have to say (working on `lines` amt), `[chunk_by](https://doc.rust-lang.org/nightly/core/primitive.slice.html#method.chunk_by)` is exactly what i need, that's really a shame. i'm on a good path to figuring out something good though

Morgane55440:
new benches for `erode()` and `dilate()` :
```
test morphology::benches::bench_grayscale_op_erode_big_image_diamond                  ... bench:   1,906,060 ns/iter (+/- 57,854)
test morphology::benches::bench_grayscale_op_erode_big_image_point                    ... bench:      71,312 ns/iter (+/- 28,722)
test morphology::benches::bench_grayscale_op_erode_medium_image_diamond               ... bench:     146,737 ns/iter (+/- 4,326)
test morphology::benches::bench_grayscale_op_erode_medium_image_large_square          ... bench:   8,119,980 ns/iter (+/- 1,010,798)
test morphology::benches::bench_grayscale_op_erode_medium_image_point                 ... bench:       4,376 ns/iter (+/- 3,639)
test morphology::benches::bench_grayscale_op_erode_small_image_diamond                ... bench:      42,703 ns/iter (+/- 3,684)
test morphology::benches::bench_grayscale_op_erode_small_image_large_square           ... bench:   1,232,370 ns/iter (+/- 361,772)
test morphology::benches::bench_grayscale_op_erode_small_image_point                  ... bench:       1,036 ns/iter (+/- 83)
```

```
test morphology::benches::bench_grayscale_op_dilate_big_image_diamond                 ... bench:   1,837,840 ns/iter (+/- 30,849)
test morphology::benches::bench_grayscale_op_dilate_big_image_point                   ... bench:      64,265 ns/iter (+/- 1,396)
test morphology::benches::bench_grayscale_op_dilate_medium_image_diamond              ... bench:     158,506 ns/iter (+/- 11,497)
test morphology::benches::bench_grayscale_op_dilate_medium_image_large_square         ... bench:   8,156,330 ns/iter (+/- 2,925,125)
test morphology::benches::bench_grayscale_op_dilate_medium_image_point                ... bench:       4,266 ns/iter (+/- 115)
test morphology::benches::bench_grayscale_op_dilate_small_image_diamond               ... bench:      41,219 ns/iter (+/- 777)
test morphology::benches::bench_grayscale_op_dilate_small_image_large_square          ... bench:   1,202,150 ns/iter (+/- 300,326)
test morphology::benches::bench_grayscale_op_dilate_small_image_point                 ... bench:       1,025 ns/iter (+/- 18)
```

Morgane55440:
ok, so it all seems good, just 2 things i was thinking about : 
- first, small and simple : i think it would be good to refactor the `erode` and `dilate` code into  a single function, something like
```per_pixel_mask_based_reduce(image: &GrayImage, mask: &Mask, neutral : u8, operator : Fn (u8,u8) -> u8)```i'd like a second opinion to be sure, especially on the name
- second, less simple, maybe not super relevant anymore : i mentioned adding parallel  versions of the user-facing functions in the beginning of the PR. that wasn't the focus at first, but i think it could easily be completely finished by tomorrow. of course, it might make more sense to leave that for another PR

**Comment by Morgane55440** [comment]
then it should all be ready to merge

**Comment by Morgane55440** [comment]
i did the refactor, if anyone has a better function name idea, i'm all ears, otherwise, i think this could be good to merge

File: src/morphology.rs [hunk 10]
```diff
@@ -361,12 +364,17 @@ pub struct Mask {
...
     /// - all the integer values will be strictly between -512 and 512
-    /// - all tuples will be sorted in lexicographic order
+    /// - all tuples will be sorted in reverse lexicographic order, line by line ((-1,-1),(0,-1),(1,-1),(-1,0),(0,0),...)
     /// - no tuple shall appear twice
-    elements: Vec<(i16, i16)>,
+    elements: Vec<Point<i16>>,
 }
 
 impl Mask {
+    fn new(elements: Vec<Point<i16>>) -> Self {
```
ripytide, line 373: Should we make this public to allow custom `Mask`s?

src/morphology.rs
```diff
@@ -576,33 +588,76 @@ impl Mask {
...
-            .map(move |(i, j)| (x as i64 + *i as i64, y as i64 + *j as i64))
-            .filter(move |(i, j)| {
-                0 <= *i && *i < image.width() as i64 && 0 <= *j && *j < image.height() as i64
-            })
-            .map(move |(i, j)| image.get_pixel(i as u32, j as u32))
+    /// all the slices are garanteed to be non empty
+    fn lines<'a>(
+        &'a self,
+    ) -> GroupBy<i16, Iter<Point<i16>>, impl FnMut(&&'a Point<i16>) -> i16 + '_> {
+        self.elements.iter().group_by(|p: &&'a Point<i16>| p.y)
```
ripytide, line 629: This seems to assume that the elements are sorted into y_groups. Should be enforce this ordering in `Mask::new()`? Or is this just for performance reasons in case points are listed in y-groups which might always be the case for the various constructors?

ripytide (commented):
Excellent work!

Morgane55440:
@ripytide 
all of the comments remaining are about `Mask::new()`.

this was not part of my intended code, i added it due to a suggestion by @cospectrum , and i think it was a mistake on my part
> You can add private `new` and call it instead of `Self { elements }` everywhere
> 
> ```rust
> fn new(elements: Vec<Point<i16>>) -> Self {
>     assert!(elements.len() <= (511 * 511) as usize);
>     Self { elements } 
> }
> ```
the code that i want to make guarantees that for all `Mask`s : 
 all the integer values will be strictly between -512 and 512
- all tuples will be sorted in lexicographic order
- all tuples will be sorted in reverse lexicographic order, line by line ((-1,-1),(0,-1),(1,-1),(-1,0),(0,0),...)
 - no tuple shall appear twice
 this shall be true for all `Mask`s, and is garanteed by the current implementation of all the user-facing constructor.
i think the `new` should just be removed. the creation of custiom maks is already possible through `from_image`

src/morphology.rs
```diff
@@ -576,33 +588,76 @@ impl Mask {
...
-        y: u32,
-    ) -> impl Iterator<Item = &'a Luma<u8>> {
-        self.elements
-            .iter()
-            .map(move |(i, j)| (x as i64 + *i as i64, y as i64 + *j as i64))
-            .filter(move |(i, j)| {
-                0 <= *i && *i < image.width() as i64 && 0 <= *j && *j < image.height() as i64
-            })
-            .map(move |(i, j)| image.get_pixel(i as u32, j as u32))
+    /// all the slices are garanteed to be non empty
```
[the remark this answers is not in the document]
Morgane55440: thanks for that

Morgane55440:
maybe i could add tests that check the conditions are always met ?

ripytide:
Ah okay, that makes sense. Since there are tests already I would expect them to fail if any of the constructors stopped abiding by the correct ordering so you probably don't need to write explicit tests for checking the ordering as that is just an implementation detail.

Morgane55440:
well then it should be good as-is

i could add a comment above the initializations, something like : 
```rust
 // direct initialization stands as guarantee that all conditions mentioned by the doc are fulfilled for the struct and each of it's parts
Self { elements }
```

Morgane55440:
thanks a ton for [typos](https://github.com/crate-ci/typos) @ripytide . i fixed the others it showed me, including some in `distance_transform` because i wrote them, they seemed relevant and there are no other PR on the subject atm

src/morphology.rs
cospectrum: No

cospectrum:
Why did you remove `new`?

Morgane55440:
@cospectrum because it didn't do anything and was confusing.

each constructor guarantees the struct it returns is valid in it's own way. 
the existence of the `new` method implied that it was validating the data. 
it wasn't, and making it truly validate the data would require it to iterate through and check the whole vector, which would slow down all the constructors just to double check something that was already assured.

Morgane55440:
replacing `Mystruct { my_data }` by `Mystruct::new(my_data )` is great, except if you're already in a constructor which is doing it's own validation

Morgane55440:
i think i have an alternative solution that may help documentation, to make clearer what's happening: 
```rust
    /// new_unchecked creates a new mask without doing any form of data validation
    /// 
    /// # Safety
    /// 
    /// this method is not user-facing and marked `unsafe` because it may lead 
    /// to invalid state if called with the wrong arguments.
    /// 
    /// By calling new_unchecked, you guarantee that :
    /// - all the integer values in `elements` will be strictly between -512 and 512
    /// - the maximum L_inf distance between any 2 points in `elements` is 512
    /// - all Points in `elements` are sorted in reverse lexicographic order, line by line ((-1,-1),(0,-1),(1,-1),(-1,0),(0,0),...)
    /// - no point appears twice in `elements`
    unsafe fn new_unchecked(elements: Vec<Point<i16>>) -> Self {
        Self { elements }
    }
```

Morgane55440:
none of it is user-facing, so it's not a huge deal, but it probably would be helpful for maintainability to have the danger laid out by the `unsafe` block and all the documentation stored in the same place

cospectrum:
> except if you're already in a constructor which is doing it's own validation

None of your constructors do real validation.
Comments prove nothing, asserts do

>the existence of the new method implied that it was validating the data. It wasn't

It does 1 check. One is better than nothing. And it may have more in the future.

>and making it truly validate the data would require it to iterate through and check the whole vector

That's why debug_assert exists

Morgane55440:
so you would prefer a `new` method with `debug_assert`s ? i can make it if you think it's the best way to guarantee safety. 

i'm just trying to make code as good as possible, following the advice and criticisms i'm given, this is all work in progress.

PR #599, opened 2024-05-08
Add `proptests` for `filter/mod.rs`

cospectrum:


ripytide:
Looks good to me. As an aside, I wonder if there are any functions in the library that are perfect inverses on one another as that would make for some really neat property tests, testing for equality after applying a function and then its inverse.

cospectrum:
There will also be more checks under sanitizers or with `debug_assert`. `Miri` is too slow for this

theotherphil:
Thanks! It’s a bit clunky to have proptest and quickcheck used in the same crate - would be nice to move the quickcheck uses across at some point.

cospectrum:
I think `proptest` is a more powerful lib, I can get rid of `quickchek`, but its in the public api behind the feature.

theotherphil:
I expect it’s not widely used and I’m fine with the API change if it’s easy for users to migrate.

cospectrum:
Why do we even have `property_testing` in the public api?

cospectrum:
It's not used in integration tests, so it doesn't make sense to me.

theotherphil:
It's in the public API to help users test their own functions. It's been used in at least one other crate but I'm not sure how widely it's used now.

Issue #520, opened 2023-05-03
Add generalizated convolution filter


vsuryamurthy:
At the moment, there is only a filter3x3 available. I would like to add a function which is a generalization of filter3x3.
Here is the PR: https://github.com/image-rs/imageproc/pull/517

willhansen:
I thought it was odd I couldn't find this feature

theotherphil:
This is available as `Kernel::filter`

ripytide:
I think one of the reasons it might not have been spotted is because it is an associated method to the `Kernel` struct which is rather inconsistent as nearly all image operations are free-standing functions.

I think making that function free-standing would be a good idea.

theotherphil:
I’d accept a PR to make that change.

PR #601, opened 2024-05-10
Gradient enum and Kernel trait V2

ripytide:
This is a version 2 PR of #596

This time I've added a `Kernel` trait to allow heap OR stack allocated kernel usage using `OwnedKernel` and `BorrowedKernel` respectively. I also rolled back the 1D kernel function back to taking `&[K]`.

Summary:

- Made more functions taking 2D kernels the `impl Kernel<K>` trait for Heap or Stack kernels
- Moved `HORIZONTAL_SOBEL` and similar constants to `OwnedKernel::sobel_horizontal_3x3()` and made them generic for any `T: From<i8>
- Removed `horizontal_*` and `vertical_*` functions since they can all be achieved via composition such as `filter(image, OwnedKernel::sobel_horizontal_3x3())` which is more explicit and rusty.
- Bumped MSRV to 1.75.0 for the `enumerate()` method of `Kernel` which uses return position impl traits.
- Made `Kernel::filter` a free-standing function which is more consistent with the rest of the library `crate::filter::filter()`
- Added a `filter_clamped()` function
- Removed the `filter3x3()` function as it is no different from `filter_clamped()`

### DIFF
--- a/examples/gradients.rs
+++ b/examples/gradients.rs
@@ -5,7 +5,7 @@
 //! `cargo run --example gradients ./examples/empire-state-building.jpg ./tmp`
 
 use image::{open, GrayImage};
-use imageproc::{filter::filter3x3, gradients::GradientKernel, map::map_subpixels};
+use imageproc::{filter::filter_clamped, gradients::GradientKernel, map::map_subpixels};
 use std::{env, fs, path::Path};
 
 fn save_gradients(
@@ -15,8 +15,8 @@ fn save_gradients(
     name: &str,
     scale: i16,
 ) {
-    let horizontal_gradients = filter3x3(input, gradient_kernel.kernel1::<i16>());
-    let vertical_gradients = filter3x3(input, gradient_kernel.kernel2::<i16>());
+    let horizontal_gradients = filter_clamped(input, gradient_kernel.kernel1::<i16>());
+    let vertical_gradients = filter_clamped(input, gradient_kernel.kernel2::<i16>());
 
     let horizontal_scaled = map_subpixels(&horizontal_gradients, |p: i16| (p.abs() / scale) as u8);
     let vertical_scaled = map_subpixels(&vertical_gradients, |p: i16| (p.abs() / scale) as u8);
--- a/examples/gradients.rs
+++ b/examples/gradients.rs
@@ -5,18 +5,19 @@
 //! `cargo run --example gradients ./examples/empire-state-building.jpg ./tmp`
 
 use image::{open, GrayImage};
-use imageproc::{filter::filter_clamped, gradients::GradientKernel, map::map_subpixels};
+use imageproc::{filter::filter_clamped, kernel::OwnedKernel, map::map_subpixels};
 use std::{env, fs, path::Path};
 
 fn save_gradients(
     input: &GrayImage,
-    gradient_kernel: &GradientKernel,
+    horizontal_kernel: &OwnedKernel<i16>,
+    vertical_kernel: &OwnedKernel<i16>,
     output_dir: &Path,
     name: &str,
     scale: i16,
 ) {
-    let horizontal_gradients = filter_clamped(input, gradient_kernel.kernel1::<i16>());
-    let vertical_gradients = filter_clamped(input, gradient_kernel.kernel2::<i16>());
+    let horizontal_gradients = filter_clamped(input, horizontal_kernel);
+    let vertical_gradients = filter_clamped(input, vertical_kernel);
 
     let horizontal_scaled = map_subpixels(&horizontal_gradients, |p: i16| (p.abs() / scale) as u8);
     let vertical_scaled = map_subpixels(&vertical_gradients, |p: i16| (p.abs() / scale) as u8);
@@ -57,19 +58,38 @@ fn main() {
     let gray_path = output_dir.join("grey.png");
     input_image.save(&gray_path).unwrap();
 
-    for gradient_kernel in GradientKernel::ALL {
-        let scale = match gradient_kernel {
-            GradientKernel::Sobel => 32,
-            GradientKernel::Scharr => 8,
-            GradientKernel::Prewitt => 6,
-            GradientKernel::Roberts => 4,
-        };
-
+    for (name, horizontal, vertical, scale) in [
+        (
+            "sobel",
+            OwnedKernel::sobel_horizontal_3x3(),
+            OwnedKernel::sobel_vertical_3x3(),
+            32,
+        ),
+        (
+            "scharr",
+            OwnedKernel::scharr_horizontal_3x3(),
+            OwnedKernel::scharr_vertical_3x3(),
+            8,
+        ),
+        (
+            "prewitt",
+            OwnedKernel::prewitt_horizontal_3x3(),
+            OwnedKernel::prewitt_vertical_3x3(),
+            6,
+        ),
+        (
+            "roberts",
+            OwnedKernel::roberts_horizontal_2x2(),
+            OwnedKernel::roberts_vertical_2x2(),
+            4,
+        ),
+    ] {
         save_gradients(
             &input_image,
-            &gradient_kernel,
+            &horizontal,
+            &vertical,
             output_dir,
-            &format!("{gradient_kernel:?}"),
+            name,
             scale,
         );
     }
--- a/examples/gradients.rs
+++ b/examples/gradients.rs
@@ -4,29 +4,29 @@
 //!
 //! `cargo run --example gradients ./examples/empire-state-building.jpg ./tmp`
 
-use image::{open, GrayImage, Luma};
-use imageproc::{
-    definitions::Image,
-    gradients::{
-        horizontal_prewitt, horizontal_scharr, horizontal_sobel, vertical_prewitt, vertical_scharr,
-        vertical_sobel,
-    },
-    map::map_pixels,
-};
+use image::{open, GrayImage};
+use imageproc::{filter::filter3x3, gradients::GradientKernel, map::map_subpixels};
 use std::{env, fs, path::Path};
 
-fn compute_gradients<F>(
+fn save_gradients(
     input: &GrayImage,
-    gradient_func: F,
+    gradient_kernel: &GradientKernel,
     output_dir: &Path,
-    file_name: &str,
+    name: &str,
     scale: i16,
-) where
-    F: Fn(&GrayImage) -> Image<Luma<i16>>,
-{
-    let gradient = gradient_func(input);
-    let gradient = map_pixels(&gradient, |_, _, p| Luma([(p[0].abs() / scale) as u8]));
-    gradient.save(&output_dir.join(file_name)).unwrap();
+) {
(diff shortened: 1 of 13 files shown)


examples/gradients.rs
```diff
@@ -4,29 +4,29 @@
...
+    name: &str,
     scale: i16,
-) where
-    F: Fn(&GrayImage) -> Image<Luma<i16>>,
-{
-    let gradient = gradient_func(input);
-    let gradient = map_pixels(&gradient, |_, _, p| Luma([(p[0].abs() / scale) as u8]));
-    gradient.save(&output_dir.join(file_name)).unwrap();
+) {
+    let horizontal_gradients = filter3x3(input, gradient_kernel.kernel1::<i16>());
```
cospectrum, line 18: I don't know why should change filter3x3 signature. Just `Kernel::filter`

cospectrum (commented):
Breaking changes, and there are only 4 days left before the release.

examples/gradients.rs
ripytide: I don't know what you mean?

ripytide:
I'm in no rush to get this merged before the next release, the breaking changes are allowable since we're in pre 1.0 stage, and justified given `gradients()` and `filter()` are much more flexible now than before and `GradientKernel` is more ergonomic in my opinion.

examples/gradients.rs
ripytide: The reason is because of #520 where users are struggling to discover the `filter()` method because it is a associated method rather than a free-standing function like every other function in the library. This is very inconsistent and so it would be better to just use `filter3x3(image, kernel)` or `filter(image, kernel)` instead of `kernel.filter(image)`.

cospectrum: I think we can do it without breaking changes!
What about `pub fn filter(image: &Image<P>, kernel: &Kernel<K>, f: impl FnMut(&mut Q::Subpixel, K)) -> Image<Q>`?
We can also implement `Deref`, so we can call `filter3x3(&img, &kernel)`
```
impl<'a, T> Deref for Kernel<'a, T> {
    type Target = [T];
    fn deref(&self) -> &Self::Target {
        self.data
    }
}
```

cospectrum:
Creating `traits` and an additional `Kernel` structs is too much for a small issue like https://github.com/image-rs/imageproc/issues/520. The owner didn't like it, but I have suggested a possible solution.

examples/gradients.rs
cospectrum: It's not difficult to find the `Kernel` methods, we just don't have examples. Also there is no `clamped_filter` method, which is easier to use

ripytide: I added a `filter_clamped()` method in this PR for convenience.

> pub fn filter<'a>(image: &Image<P>, kernel: &Kernel<'a, K>, f: impl FnMut(&mut Q::Subpixel, K)) -> Image<Q>

This is already possible by the `filter()` method in this PR

ripytide:
The point is that consistency is important and having some important methods be free-standing and other associated with types is inconsistent. This PR fixes that. The traits are simply much more flexible as they allow heap OR stack allocated kernels which is necessary for dynamically sized kernels such as for the `GradientKernel` which is 3x3 or 2x2.

examples/gradients.rs
cospectrum: But you have trait

ripytide: Yes the trait is used in the `filter_clamped()` function. Its signature is simply: `pub fn filter_clamped<P, K, S>(image: &Image<P>, kernel: impl Kernel<K>) -> Image<ChannelMap<P, S>>`

cospectrum:
> which is necessary for dynamically sized kernels such as for the GradientKernel which is 3x3 or 2x2

No, it's not
```rust
#[derive(Debug, Copy, Clone)]
pub enum Gradient {
    /// The 3x3 Sobel kernel
    /// See <https://en.wikipedia.org/wiki/Sobel_operator>
    Sobel,
    /// The 3x3 Scharr kernel
    Scharr,
    /// The 3x3 Prewitt kernel
    Prewitt,
    /// The 2x2 Roberts kernel
    /// See <https://en.wikipedia.org/wiki/Roberts_cross>
    Roberts,
}

impl Gradient {
    pub fn vertical_kernel(&self) -> Kernel<'static, i32> {
        match self {
            Gradient::Sobel => Kernel::new(&VERTICAL_SOBEL, 3, 3),
            Gradient::Scharr => Kernel::new(&VERTICAL_SCHARR, 3, 3),
            Gradient::Prewitt => Kernel::new(&VERTICAL_PREWITT, 3, 3),
            Gradient::Roberts => Kernel::new(&VERTICAL_ROBERTS, 2, 2),
        }
    }
}
```

examples/gradients.rs
ripytide: `OwnedKernel` and `BorrowedKernel` are the main two types which now implement the `Kernel` trait.

ripytide:
Ok, it's not necessary for the `GradientKernel` enum, but it is necessary for run-time dynamically sized kernels, like for example a cli script that took the kernel size as a command-line argument and then did a convolution based on that kernel size.

How would you implement this?

```rust
pub fn guass_kernel(radius: u32) -> BorrowedKernel<'static, i32> {
    //impossible without leaking memory
}
```

cospectrum:
I would just return `Vec<i32>` or `Image<Luma<T>>`.
Maybe you should create an issue and ask @theotherphil about some changes in `Kernel` api? This way you won't waste your time writing code that won't be accepted.

ripytide:
`Vec<i32>` is missing dimensions, `Image` cannot be used with any of the `filter` functions both are inferior to `OwnedKernel` + the `Kernel` trait.

A PR is easier to discuss changes over than an issue as it shows the full extent of a change required.

cospectrum:
Maybe we can add `from_image` and that's it.
```rs
impl<'a, T> Kernel<'a, T>
where T: Primitive
{
    /// Create kernel from image
    pub fn from_image(image: &'a Image<Luma<T>>) -> Self {
        Self::new(image.as_ref(), image.width(), image.height())
    }
}
```

ripytide:
But that still doesn't allow the implementation of `guass_kernel()`.
```rust
pub fn guass_kernel(radius: u32) -> BorrowedKernel<'static, i32> {
    let image = (0..radius).cartesian_product(0..radius);
    BorrowedKernel::from_image(image) // borrow checker issue
}
```

cospectrum:
So, the question is whether we want struct `OwnedKernel` and `Filter` trait or not.

ripytide:
It adds flexibility to the library to do things not previously possible like my cli kernel example. Why not add it @cospectrum?

cospectrum:
I don't have any strong arguments for or against. your example make sense, but there is one point: in the library itself there is not yet a single place where we could use this `OwnedKernel`. That is, it will be an export-only struct.

theotherphil:
Thanks for your work on exploring options to improve filter ergonomics.

Unfortunately these changes add too much user facing complexity for the benefits they bring.

Could the discoverability issue of `Kernel::filter` be fixed by just making this associated function freestanding?

ripytide:
Could the complexity of using a `Kernel` trait can be mitigated by adding examples to methods taking kernel arguments? I disagree that the ergonomics and flexibility is not worth the complexity.

Dynamically sized kernels seems to me to be a fairly reasonable usecase, for example, `opencv` has the `ksize` parameter (https://docs.opencv.org/4.x/d4/d86/group__imgproc__filter.html#gaabe8c836e97159a9193fb0b11ac52cf1) which would be made much easier to implement using the `OwnedKernel`.

cospectrum:
I don't really understand what the problem is with the `Kernel::filter'. People saw `filter3x3` function and wanted `filter_nxm`, they didn't want the `Kernel trait`. And the `trait` will cause much more trouble to the user than just a structure with a method.

ripytide:
Wait a minute, @cospectrum
> No, it's not
```rust
#[derive(Debug, Copy, Clone)]
pub enum Gradient {
    /// The 3x3 Sobel kernel
    /// See <https://en.wikipedia.org/wiki/Sobel_operator>
    Sobel,
    /// The 3x3 Scharr kernel
    Scharr,
    /// The 3x3 Prewitt kernel
    Prewitt,
    /// The 2x2 Roberts kernel
    /// See <https://en.wikipedia.org/wiki/Roberts_cross>
    Roberts,
}

impl Gradient {
    pub fn vertical_kernel(&self) -> Kernel<'static, i32> {
        match self {
            Gradient::Sobel => Kernel::new(&VERTICAL_SOBEL, 3, 3),
            Gradient::Scharr => Kernel::new(&VERTICAL_SCHARR, 3, 3),
            Gradient::Prewitt => Kernel::new(&VERTICAL_PREWITT, 3, 3),
            Gradient::Roberts => Kernel::new(&VERTICAL_ROBERTS, 2, 2),
        }
    }
}
```

This doesn't work for any type T whereas my implementation works for any `T: From<i8>`

ripytide:
You would have to write a constant for every single type which would be super annoying to use.

ripytide:
I've now removed the `enum` and moved all the Kernel constructors to associated methods of `OwnedKernel`.

cospectrum:
> Wait a minute, @cospectrum
> 
> > No, it's not
> 
> ```rust
> #[derive(Debug, Copy, Clone)]
> pub enum Gradient {
>     /// The 3x3 Sobel kernel
>     /// See <https://en.wikipedia.org/wiki/Sobel_operator>
>     Sobel,
>     /// The 3x3 Scharr kernel
>     Scharr,
>     /// The 3x3 Prewitt kernel
>     Prewitt,
>     /// The 2x2 Roberts kernel
>     /// See <https://en.wikipedia.org/wiki/Roberts_cross>
>     Roberts,
> }
> 
> impl Gradient {
>     pub fn vertical_kernel(&self) -> Kernel<'static, i32> {
>         match self {
>             Gradient::Sobel => Kernel::new(&VERTICAL_SOBEL, 3, 3),
>             Gradient::Scharr => Kernel::new(&VERTICAL_SCHARR, 3, 3),
>             Gradient::Prewitt => Kernel::new(&VERTICAL_PREWITT, 3, 3),
>             Gradient::Roberts => Kernel::new(&VERTICAL_ROBERTS, 2, 2),
>         }
>     }
> }
> ```
> 
> This doesn't work for any type T, only i32, whereas my implementation works for any `T: From<i8>`

But we don't need any type.

ripytide:
Not yet, but what if users have different image pixel formats: `u8`, `i8`, `u16`, etc.. it would be very wasteful to always cast to `i32`. Would it not be better if the kernels could be created in any generic format from `i8`?

cospectrum:
A dynamic cast from one type to another will still be a waste of resources. I would expect more from the library. For example, generate all static arrays using macros.

ripytide:
Despite the fact that kernel creation is incomparably tiny when compared with image convolution? That seems like a heck of performance/complexity tradeoff. A lot of the casts can also be optimized by the compiler since the generic will always be known at compile time.

ripytide:
I'm going to close this and work on a V3 that discards the disagreeable things like the Kernel trait and just focuses on the minimal other things that have been largely agreed on.

Issue #7, opened 2015-10-02
Integration with image


theotherphil:
When this library is sufficiently well featured and stable we should remove the overlap between it and the image crate. Everything in image::imagops looks to naturally belong here, but before doing that we need to:
1. find uses of those operations in other crates (this is hopefully as easy as just using cargo-crusader),
2. get agreement from the authors of the image crate!
3. decide how to handle DynamicImage. This currently forwards each of the stable-and-fairly-short list of imageops functions to those on the corresponding GenericImage. It's not sensible to just keep growing this list forever, but the simplest use case of load -> op -> save should remain as simple as possible.

ripytide:
I've opened https://github.com/image-rs/image/issues/2238 and the maintainer seems open to the idea of migrating functionality on a case-by-case basis.

My current action plan is to individually migrate all/most of the functionality from `imageops` to `imageprocs` starting with the most `imageprocs` relevant functions. After each function is integrated into `imageprocs` then the `image` crate maintainers can then deprecate the functions which would allow them to remove them at some point in the future after users have hopefully migrated.

How does that sound?

> decide how to handle DynamicImage

Looking through the `imageops` code it seems most functions are now simply using `GenericImageView` traits exactly how most current `imageprocs` functions are implemented which might make things much easier than they might have been previously.

theotherphil:
Let's consolidate discussion on the image crate issue - I'll close this one.

PR #602, opened 2024-05-12
Parallel Pixel Maps

ripytide:
This is just a draft so I can get some feedback on an initial implementation of `map_subpixels_parallel()` before I add the remaining variation of functions `map_subpixels_parallel_mut()` `map_pixels_parallel()` `map_pixels_parallel_mut()`.

This would address #429

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -1,4 +1,4 @@
-//! Functions for mapping over pixels, colors or subpixels of images.
+//! Functions for mapping pixels and subpixels of images.
 
 use image::{GenericImage, Luma, LumaA, Pixel, Primitive, Rgb, Rgba};
 
@@ -182,7 +182,7 @@ where
     }
 }
 
-/// Applies `f` to the color of each pixel in the input image.
+/// Applies `f` to each pixel of the input image.
 ///
 /// # Examples
 /// ```
@@ -191,7 +191,7 @@ where
 /// # extern crate imageproc;
 /// # fn main() {
 /// use image::Rgb;
-/// use imageproc::map::map_colors;
+/// use imageproc::map::map_pixels;
 ///
 /// let image = gray_image!(
 ///     1, 2;
@@ -202,11 +202,11 @@ where
 ///     [3, 6, 9], [4, 8, 12]);
 ///
 /// assert_pixels_eq!(
-///     map_colors(&image, |p| { Rgb([p[0], (2 * p[0]), (3 * p[0])]) }),
+///     map_pixels(&image, |p| { Rgb([p[0], (2 * p[0]), (3 * p[0])]) }),
 ///     rgb);
 /// # }
 /// ```
-pub fn map_colors<I, P, Q, F>(image: &I, f: F) -> Image<Q>
+pub fn map_pixels<I, P, Q, F>(image: &I, f: F) -> Image<Q>
 where
     I: GenericImage<Pixel = P>,
     P: Pixel,
@@ -228,7 +228,7 @@ where
     out
 }
 
-/// Applies `f` to the color of each pixel in the input image in place.
+/// Applies `f` to each pixel of the input image in place.
 ///
 /// # Examples
 /// ```
@@ -237,7 +237,7 @@ where
 /// # extern crate imageproc;
 /// # fn main() {
 /// use image::Luma;
-/// use imageproc::map::map_colors_mut;
+/// use imageproc::map::map_pixels_mut;
 ///
 /// let mut image = gray_image!(
 ///     1, 2;
@@ -247,14 +247,14 @@ where
 ///     2, 4;
 ///     6, 8);
 ///
-/// map_colors_mut(&mut image, |p| Luma([2 * p[0]]));
+/// map_pixels_mut(&mut image, |p| Luma([2 * p[0]]));
 ///
 /// assert_pixels_eq!(
 ///     image,
 ///     want);
 /// # }
 /// ```
-pub fn map_colors_mut<I, P, F>(image: &mut I, f: F)
+pub fn map_pixels_mut<I, P, F>(image: &mut I, f: F)
 where
     I: GenericImage<Pixel = P>,
     P: Pixel,
@@ -272,9 +272,12 @@ where
     }
 }
 
-/// Applies `f` to the colors of the pixels in the input images.
+/// Applies `f` to each pixel of both input images.
 ///
-/// Requires `image1` and `image2` to have the same dimensions.
+/// # Panics
+///
+/// Panics if `image1` and `image2` do not have the same dimensions.
+///
 /// # Examples
 /// ```
 /// # extern crate image;
@@ -282,7 +285,7 @@ where
 /// # extern crate imageproc;
 /// # fn main() {
 /// use image::Luma;
-/// use imageproc::map::map_colors2;
+/// use imageproc::map::map_pixels2;
 ///
 /// let image1 = gray_image!(
 ///     1, 2,
@@ -300,12 +303,12 @@ where
 /// );
 ///
 /// assert_pixels_eq!(
-///     map_colors2(&image1, &image2, |p, q| Luma([p[0] + q[0]])),
+///     map_pixels2(&image1, &image2, |p, q| Luma([p[0] + q[0]])),
 ///     sum
 /// );
 /// # }
 /// ```
-pub fn map_colors2<I, J, P, Q, R, F>(image1: &I, image2: &J, f: F) -> Image<R>
+pub fn map_pixels2<I, J, P, Q, R, F>(image1: &I, image2: &J, f: F) -> Image<R>
 where
     I: GenericImage<Pixel = P>,
     J: GenericImage<Pixel = Q>,
@@ -332,7 +335,7 @@ where
     out
 }
 
-/// Applies `f` to each pixel in the input image.
+/// Applies `f` to each enumerated pixel of the input image.
 ///
 /// # Examples
 /// ```
@@ -341,7 +344,7 @@ where
 /// # extern crate imageproc;
 /// # fn main() {
 /// use image::Rgb;
-/// use imageproc::map::map_pixels;
+/// use imageproc::map::map_enumerated_pixels;
 ///
 /// let image = gray_image!(
 ///     1, 2;
@@ -352,13 +355,13 @@ where
 ///     [3, 0, 1], [4, 1, 1]);
 ///
 /// assert_pixels_eq!(
-///     map_pixels(&image, |x, y, p| {
+///     map_enumerated_pixels(&image, |x, y, p| {
 ///         Rgb([p[0], x as u8, y as u8])
 ///     }),
 ///     rgb);
 /// # }
 /// ```
-pub fn map_pixels<I, P, Q, F>(image: &I, f: F) -> Image<Q>
+pub fn map_enumerated_pixels<I, P, Q, F>(image: &I, f: F) -> Image<Q>
 where
     I: GenericImage<Pixel = P>,
     P: Pixel,
@@ -380,7 +383,7 @@ where
     out
 }
 
-/// Applies `f` to each pixel in the input image in place.
+/// Applies `f` to each enumerated pixel of the input image in plpace.
 ///
 /// # Examples
 /// ```
@@ -389,7 +392,7 @@ where
 /// # extern crate imageproc;
 /// # fn main() {
 /// use image::Luma;
-/// use imageproc::map::map_pixels_mut;
+/// use imageproc::map::map_enumerated_pixels_mut;
 ///
 /// let mut image = gray_image!(
 ///     1, 2;
@@ -399,7 +402,7 @@ where
 ///     1, 3;
 ///     4, 6);
 ///
-/// map_pixels_mut(&mut image, |x, y, p| {
+/// map_enumerated_pixels_mut(&mut image, |x, y, p| {
 ///     Luma([p[0] + x as u8 + y as u8])
 /// });
 ///
@@ -408,7 +411,7 @@ where
 ///     want);
 /// # }
 /// ```
-pub fn map_pixels_mut<I, P, F>(image: &mut I, f: F)
+pub fn map_enumerated_pixels_mut<I, P, F>(image: &mut I, f: F)
 where
     I: GenericImage<Pixel = P>,
     P: Pixel,
@@ -455,7 +458,7 @@ where
     Rgb<C>: Pixel<Subpixel = C>,
     C: Primitive,
 {
-    map_colors(image, |p| Luma([p[0]]))
+    map_pixels(image, |p| Luma([p[0]]))
 }
 
 /// Creates an RGB image by embedding a grayscale image in its red channel.
@@ -487,7 +490,7 @@ where
     Rgb<C>: Pixel<Subpixel = C>,
     C: Primitive,
 {
-    map_colors(image, |p| {
+    map_pixels(image, |p| {
         let mut cs = [C::zero(); 3];
         cs[0] = p[0];
         Rgb(cs)
@@ -523,7 +526,7 @@ where
     Rgb<C>: Pixel<Subpixel = C>,
     C: Primitive,
 {
-    map_colors(image, |p| Luma([p[1]]))
+    map_pixels(image, |p| Luma([p[1]]))
 }
 
 /// Creates an RGB image by embedding a grayscale image in its green channel.
@@ -555,7 +558,7 @@ where
     Rgb<C>: Pixel<Subpixel = C>,
     C: Primitive,
 {
-    map_colors(image, |p| {
+    map_pixels(image, |p| {
         let mut cs = [C::zero(); 3];
         cs[1] = p[0];
         Rgb(cs)
@@ -591,7 +594,7 @@ where
     Rgb<C>: Pixel<Subpixel = C>,
     C: Primitive,
 {
-    map_colors(image, |p| Luma([p[2]]))
+    map_pixels(image, |p| Luma([p[2]]))
 }
 
 /// Creates an RGB image by embedding a grayscale image in its blue channel.
@@ -623,7 +626,7 @@ where
     Rgb<C>: Pixel<Subpixel = C>,
     C: Primitive,
 {
-    map_colors(image, |p| {
+    map_pixels(image, |p| {
         let mut cs = [C::zero(); 3];
         cs[2] = p[0];
         Rgb(cs)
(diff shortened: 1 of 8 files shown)


ripytide:
I've just realized this was attempted in #446 but was closed due to being slower than the original. After reading the PR it looks like it used `par_bridge` as `image` may not have had raw `par` iterators 4 years ago which it now has. I'll do some benchmarks to test how my implementation is compared to the non-parallel version.

ripytide:
Here are the results:
```

running 4 tests
test map::benches::bench_map_subpixels_100           ... bench:       9,856 ns/iter (+/- 1,715)
test map::benches::bench_map_subpixels_1000              ... bench:   1,047,116 ns/iter (+/- 42,256)
test map::benches::bench_map_subpixels_parallel_100   ... bench:      26,220 ns/iter (+/- 14,109)
test map::benches::bench_map_subpixels_parallel_1000   ... bench:     375,886 ns/iter (+/- 170,319)
```
As can be seen which function is faster depends on the size of the image, 100x100 images is faster with the non-parallel version whereas the 1000x1000 image is faster with the parallel version.

Morgane55440:
my PC is slower than yours (probably all the other stuff i'm running), but i might have something to help for the inefficiencies for low values.

lots of time is lost on dispatching, so we need to dispatch less, by parallelizing on lines instead of pixels. 
of course, there is no support for this from the image crate, so i had to improvize.

here are my results : 
your impl on my pc
```
test map::benches::bench_map_subpixels_100                                            ... bench:      16,226 ns/iter (+/- 189)
test map::benches::bench_map_subpixels_1000                                           ... bench:   2,021,782 ns/iter (+/- 138,602)
test map::benches::bench_map_subpixels_parallel_100                                   ... bench:      31,478 ns/iter (+/- 3,454)
test map::benches::bench_map_subpixels_parallel_1000                                  ... bench:     685,178 ns/iter (+/- 105,519)
```
my impl on my pc
```
test map::benches::bench_map_subpixels_100                                            ... bench:      16,320 ns/iter (+/- 214)
test map::benches::bench_map_subpixels_1000                                           ... bench:   2,001,935 ns/iter (+/- 114,785)
test map::benches::bench_map_subpixels_parallel_100                                   ... bench:      15,008 ns/iter (+/- 2,176)
test map::benches::bench_map_subpixels_parallel_1000                                  ... bench:     634,393 ns/iter (+/- 158,002)
```
it is still much slower for very small images, so one could just default to sequential for images under 10k pixels
```
test map::benches::bench_map_subpixels_10                                             ... bench:         200 ns/iter (+/- 12)
test map::benches::bench_map_subpixels_100                                            ... bench:      16,187 ns/iter (+/- 278)
test map::benches::bench_map_subpixels_1000                                           ... bench:   2,009,675 ns/iter (+/- 115,502)
test map::benches::bench_map_subpixels_parallel_10                                    ... bench:       5,757 ns/iter (+/- 1,303)
test map::benches::bench_map_subpixels_parallel_100                                   ... bench:      15,283 ns/iter (+/- 1,232)
test map::benches::bench_map_subpixels_parallel_1000                                  ... bench:     699,168 ns/iter (+/- 148,572)
```

Morgane55440:
here is my take on the function body

```
{
    use itertools::Itertools;
    use rayon::iter::IndexedParallelIterator;
    use rayon::iter::ParallelIterator;
    use rayon::slice::{ParallelSlice, ParallelSliceMut};

    let mut out: ImageBuffer<ChannelMap<P, S>, Vec<S>> = ImageBuffer::new(width, height);

    image
        .par_chunks(width as usize * <P as Pixel>::CHANNEL_COUNT as usize)
        .zip_eq(
            out.par_chunks_mut(
                width as usize * <ChannelMap<P, S> as Pixel>::CHANNEL_COUNT as usize,
            ),
        )
        .for_each(|(input_line, output_line)| {
            input_line
                .chunks_exact(<P as Pixel>::CHANNEL_COUNT as usize)
                .map(|v| <P as Pixel>::from_slice(v))
                .zip_eq(
                    output_line
                        .chunks_exact_mut(<ChannelMap<P, S> as Pixel>::CHANNEL_COUNT as usize)
                        .map(|v| <ChannelMap<P, S> as Pixel>::from_slice_mut(v)),
                )
                .for_each(|(input_pixel, output_channels)| {
                    input_pixel
                        .channels()
                        .iter()
                        .zip(output_channels.channels_mut())
                        .for_each(|(input_subpixel, output_subpixel)| {
                            *output_subpixel = f(*input_subpixel)
                        })
                });
        });

    out
}
```
this is more efficient the faster the function is, which is great when it's just a sum or multiplication

ripytide:
Great work, line based parallelism does seem like the best option for CPU cache since all the memory would be right next to each other. I wonder how these scale at 10000x10000.

Morgane55440:
i have a better iteration, that is both faster for small values and more readable

```rust
{
    use rayon::iter::IndexedParallelIterator;
    use rayon::iter::ParallelIterator;
    use rayon::slice::{ParallelSlice, ParallelSliceMut};

    let mut out: ImageBuffer<ChannelMap<P, S>, Vec<S>> = ImageBuffer::new(width, height);

    const SUBPIXEL_PER_CHUNK : usize = 512;
    image
        .par_chunks(SUBPIXEL_PER_CHUNK)
        .zip_eq(out.par_chunks_mut(SUBPIXEL_PER_CHUNK))
        .for_each(|(input_line, output_line)| {
            input_line.iter().zip(output_line.iter_mut()).for_each(
                |(input_subpixel, output_subpixel)| *output_subpixel = f(*input_subpixel),
            );
        });

    out
}
```

Morgane55440:
`SUBPIXEL_PER_CHUNK` is arbitrary, but 512 seems pretty good eperimentally

ripytide:
I've added you to my fork in case you want to update the implementation as we find faster versions.

Morgane55440:
> Great work, line based parallelism does seem like the best option for CPU cache since all the memory would be right next to each other. I wonder how these scale at 10000x10000.

sadly for me it doesn't seem to improve much beyond 3x
```
test map::benches::bench_map_subpixels_10                                             ... bench:         204 ns/iter (+/- 2)
test map::benches::bench_map_subpixels_100                                            ... bench:      16,439 ns/iter (+/- 195)
test map::benches::bench_map_subpixels_1000                                           ... bench:   2,079,892 ns/iter (+/- 395,519)
test map::benches::bench_map_subpixels_10_000                                         ... bench: 222,443,410 ns/iter (+/- 20,580,424)
test map::benches::bench_map_subpixels_parallel_10                                    ... bench:          66 ns/iter (+/- 1)
test map::benches::bench_map_subpixels_parallel_100                                   ... bench:      12,425 ns/iter (+/- 3,846)
test map::benches::bench_map_subpixels_parallel_1000                                  ... bench:     700,052 ns/iter (+/- 121,973)
test map::benches::bench_map_subpixels_parallel_10_000                                ... bench:  61,087,340 ns/iter (+/- 5,226,359)
```

Morgane55440:
the values i get from my tests are really weird, i think there might be something weird going on. 

it'd be nice if you could test it on your machine.

Morgane55440:
it says it's 10x faster for a 30 by 30 but only about 3x faster for a 1000 by 1000
```
test map::benches::bench_map_subpixels_10                                             ... bench:         192 ns/iter (+/- 1)
test map::benches::bench_map_subpixels_100                                            ... bench:      16,343 ns/iter (+/- 233)
test map::benches::bench_map_subpixels_1000                                           ... bench:   2,019,145 ns/iter (+/- 997,715)
test map::benches::bench_map_subpixels_30                                             ... bench:       1,425 ns/iter (+/- 47)
test map::benches::bench_map_subpixels_60                                             ... bench:       5,580 ns/iter (+/- 86)
test map::benches::bench_map_subpixels_parallel_10                                    ... bench:          65 ns/iter (+/- 0)
test map::benches::bench_map_subpixels_parallel_100                                   ... bench:      13,283 ns/iter (+/- 1,725)
test map::benches::bench_map_subpixels_parallel_1000                                  ... bench:     728,398 ns/iter (+/- 118,186)
test map::benches::bench_map_subpixels_parallel_30                                    ... bench:         156 ns/iter (+/- 4)
test map::benches::bench_map_subpixels_parallel_60                                    ... bench:       2,558 ns/iter (+/- 1,018)
```

Morgane55440:
small question : why does `map_subpixels_parallel` only support `ImageBuffer` while all the functions support `GenericImage` ?

ripytide:
`GenericImage` doesn't require the image to be stored contiguously, whereas `ImageBuffer` does since it implements `Deref<Target=[P::Subpixel]>` which is needed for `rayon` methods like `par_chunks()`.

Morgane55440:
well, i have bad news
all the speed increase is solely due to the fact that we're working with `ImageBuffer` and contiguous memory, and removing the parralel bits makes it all faster including 1000 by 1000

ripytide:
The benches on my machine:
```
test map::benches::bench_map_subpixels_10                                             ... bench:         207 ns/iter (+/- 2)
test map::benches::bench_map_subpixels_100                                            ... bench:      16,271 ns/iter (+/- 145)
test map::benches::bench_map_subpixels_1000                                           ... bench:   1,725,553 ns/iter (+/- 107,609)
test map::benches::bench_map_subpixels_30                                             ... bench:       1,467 ns/iter (+/- 12)
test map::benches::bench_map_subpixels_60                                             ... bench:       5,660 ns/iter (+/- 26)
test map::benches::bench_map_subpixels_parallel_10                                    ... bench:          81 ns/iter (+/- 1)
test map::benches::bench_map_subpixels_parallel_100                                   ... bench:       9,735 ns/iter (+/- 1,870)
test map::benches::bench_map_subpixels_parallel_1000                                  ... bench:     243,560 ns/iter (+/- 43,263)
test map::benches::bench_map_subpixels_parallel_30                                    ... bench:         168 ns/iter (+/- 1)
test map::benches::bench_map_subpixels_parallel_60                                    ... bench:       7,178 ns/iter (+/- 1,279)
```

Morgane55440:
what happens if you replace the body by

```rust
{
    use itertools::Itertools;

    let mut out: ImageBuffer<ChannelMap<P, S>, Vec<S>> = ImageBuffer::new(width, height);

    image.iter().zip_eq(out.iter_mut()).for_each(
            |(input_subpixel, output_subpixel)| *output_subpixel = f(*input_subpixel),
        );

    out
}


```
?

ripytide:
```
test map::benches::bench_map_subpixels_10                                             ... bench:         208 ns/iter (+/- 2)
test map::benches::bench_map_subpixels_100                                            ... bench:      16,292 ns/iter (+/- 337)
test map::benches::bench_map_subpixels_1000                                           ... bench:   1,698,793 ns/iter (+/- 202,548)
test map::benches::bench_map_subpixels_30                                             ... bench:       1,464 ns/iter (+/- 19)
test map::benches::bench_map_subpixels_60                                             ... bench:       6,032 ns/iter (+/- 54)
test map::benches::bench_map_subpixels_parallel_10                                    ... bench:          58 ns/iter (+/- 0)
test map::benches::bench_map_subpixels_parallel_100                                   ... bench:       1,035 ns/iter (+/- 9)
test map::benches::bench_map_subpixels_parallel_1000                                  ... bench:     384,083 ns/iter (+/- 22,183)
test map::benches::bench_map_subpixels_parallel_30                                    ... bench:         137 ns/iter (+/- 1)
test map::benches::bench_map_subpixels_parallel_60                                    ... bench:         393 ns/iter (+/- 2)
```

Morgane55440:
well it seems the parallel version is slightly faster for big images

ripytide:
Well at least we've learnt that there is massive room for optimization via `ImageBuffer` over `GenericImage` hehe.

Morgane55440:
still a bit perplexed at how inefficient the parallelization is, there might be something else to it i'm missing

ripytide:
From the cargo docs:
```
The --jobs argument affects the building of the benchmark executable but does not affect how many threads are used when running the benchmarks. The Rust test harness runs benchmarks serially in a single thread.
```
I think `libtest` may only be running programs with singlethreaded?

Morgane55440:
the parallelization is happening for sure

fishy_stuff

this is all very fishy

Morgane55440:
here is what happen if i remove the `with_min_len` (dramatic slow down for small images), but use more expensive computation as the input function) : 
```
test map::benches::bench_map_subpixels_parallel_10                                    ... bench:      15,389 ns/iter (+/- 2,866)
test map::benches::bench_map_subpixels_parallel_100                                   ... bench:      40,917 ns/iter (+/- 5,783)
test map::benches::bench_map_subpixels_parallel_1000                                  ... bench:     911,286 ns/iter (+/- 100,335)
test map::benches::bench_map_subpixels_parallel_10_000                                ... bench:  79,796,480 ns/iter (+/- 8,970,528)
test map::benches::bench_map_subpixels_parallel_30                                    ... bench:      26,007 ns/iter (+/- 7,937)
test map::benches::bench_map_subpixels_parallel_60                                    ... bench:      33,506 ns/iter (+/- 7,006)
test map::benches::bench_map_subpixels_sequential_10                                  ... bench:         401 ns/iter (+/- 6)
test map::benches::bench_map_subpixels_sequential_100                                 ... bench:      37,682 ns/iter (+/- 605)
test map::benches::bench_map_subpixels_sequential_1000                                ... bench:   4,110,240 ns/iter (+/- 190,140)
test map::benches::bench_map_subpixels_sequential_10_000                              ... bench: 427,149,130 ns/iter (+/- 326,256,736)
test map::benches::bench_map_subpixels_sequential_30                                  ... bench:       3,537 ns/iter (+/- 572)
test map::benches::bench_map_subpixels_sequential_60                                  ... bench:      13,524 ns/iter (+/- 326)
```

the good value of `MIN_SUBPIXEL_PER_CHUNK` completely depends on the input functions ; if we had a function that took 1ms to compute, then the parallel version would be ***much*** faster even for a 10 by 10 image.
therefore, imo, we should ship the parralel version without using `with_min_len`, explaining that it is slower than sequential for small images and fast functions, but can be significantly faster for large images(1 million pixel+) and slow fuctions

Morgane55440:
i have also come to realize that chunking manually is actually pointless, because that's already what rayon does

Morgane55440:
T_T

i thought i had done clippy on my machine

ripytide:
I've de-duplicated the benchmark mapping function into the macro and switched `map_subpixels_parallel()` to be more similar to the rest of the mapping functions using `GenericImage` but with a `Deref<[]>` bound. I tested benchmarks before and after the commits and it had no effects as expected.

ripytide:
The benchmarks take forever (200s) on my machine mainly due to the 10_000 functions (especially the sequential one), we'll probably have to remove those for the final PR.

Morgane55440:
makes sense

theotherphil:
Thanks for the PR. I’ll wait until after the next release to merge this. Which will probably be next weekend as I don’t have much time on weekdays.

src/map.rs
```diff
@@ -101,6 +101,64 @@ where
...
+    S: Primitive,
+    F: Fn(P::Subpixel) -> S + std::marker::Sync,
+    P::Subpixel: std::marker::Send,
+    P::Subpixel: std::marker::Sync,
+    P: std::marker::Send,
+    P: std::marker::Sync,
+    S: std::marker::Send,
+    S: std::marker::Sync,
+    <P as WithChannel<S>>::Pixel: std::marker::Send,
+    <P as WithChannel<S>>::Pixel: std::marker::Sync,
```
theotherphil, line 147: This `where` clause is horrific! Is there any way of shortening it?

Morgane55440: we could add `use std::marker::{Send, Sync}` at the top of the file and then
```
    S: std::marker::Send,
    S: std::marker::Sync,
```
would become
```
    S: Send + Sync,
```

Morgane55440: the final result would be something like that : 
```
pub fn map_subpixels_parallel<I, P, F, S>(image: &I, f: F) -> Image<ChannelMap<P, S>>
where
    I: GenericImage<Pixel = P> + Deref<Target = [P::Subpixel]>,
    P: WithChannel<S> + Send + Sync,
    S: Primitive + Send + Sync,
    F: Fn(P::Subpixel) -> S + Sync,
    P::Subpixel: Send + Sync,
    <P as WithChannel<S>>::Pixel: Send + Sync,
```

cospectrum: `Send` and `Sync` are already in rust prelude

cospectrum:
I think all such functions should be in the `image` crate.

ripytide:
> I think all such functions should be in the `image` crate.

Interesting point. My option would be that anything that mutates or transforms images counts as image processing and so belongs in the `imageproc` crate whereas the `image` crate is for image representation, encoding and decoding.

Since these are mapping operations "process" images and so are quite "heavy" functions they should belong in the `imageproc` crate.

That said the line here is quite blurry. I recently learned that `image` has a `imageops` module with several image processing operations such as `contrast()` and `brighten()` which is quite unsettling as I would think they belong in the `imageproc` crate.

theotherphil:
The image crate existed before imageproc. Sorting out what goes where was one of the first issues on this repo… and we’ve pretty much just ignored it for the last nine years! https://github.com/image-rs/imageproc/issues/7

theotherphil:
@ripytide this looks reasonable to me - let me know when it's in a reviewable state.

ripytide:
I think we should talk through #622 first since the outcome of that will decide how many of the function variations to implement here.

ripytide:
Okay now that all the blockers have been resolved I've fleshed out the rest of this PR.
Some other changes I made:
- Renamed `red_channel()` to `into_red_channel()` since that is more clear on the operation being done.
- Renamed `as_red_channel()` to `from_red_channel()` for the same reason as above, also `as_` functions are supposed to be cheap operations according to the rust API guidelines.
- All signatures have changes from `map_pixels<I, P>(image: &I) where I: GenericImage<P>` to `map_pixels<P>(image: &Image<P>)` since it provides substantial performance improvements as it enforces contiguous storage. This is more consistent with other `imageproc` functions which use `Image<P>`'s and is more inline with the crate principle of performance over maximum genericness.

ripytide:
@theotherphil this is now ready for review.

theotherphil:
The new code looks good and the signature change makes sense. I’m not convinced by the name changes though. You’re right that `as_` should be cheap, but there’s also a standard convention that `into_` and `from_` consume an owned value which isn’t the case here.

ripytide:
That's true, but it is the lesser evil in my opinion as users can see their signatures fairly easily from the docs. Unless you have an alternative suggestion?

ripytide:
One alternative:
`greyscale_from_red_channel()` and `rgb_from_red_grayscale()`?

theotherphil:
I’ll merge as is and think about what best to name these before the next release.

Thanks for the new functions, and it’s good to see your new doc macros already paying off.

PR #603, opened 2024-05-14
Add proptests for `Mask`

cospectrum:


src/morphology.rs
```diff
@@ -430,12 +430,25 @@ impl Mask {
...
+
+    fn new(elements: Vec<Point<i16>>) -> Self {
+        assert!(elements.len() <= (511 * 511) as usize);
+        debug_assert!({
+            let unique = elements
+                .iter()
+                .map(|p| (p.x, p.y))
+                .collect::<std::collections::HashSet<_>>();
+            unique.len() == elements.len()
+        });
```
ripytide, line 451:
```suggestion
        debug_assert!(elements.iter().all_unique());
```
Use `Itertools` to make this more readable.

src/point.rs
```diff
@@ -59,6 +59,11 @@ impl<T: NumCast> Point<T> {
     pub(crate) fn to_i32(&self) -> Point<i32> {
         Point::new(self.x.to_i32().unwrap(), self.y.to_i32().unwrap())
     }
+
+    /// Converts to a Point<i32>. Panics if the cast fails.
```
ripytide, line 63:
```suggestion
    /// Converts to a Point<i16>. Panics if the cast fails.
```

src/morphology.rs
```diff
@@ -430,12 +430,25 @@ impl Mask {
...
         Self { elements }
```
ripytide, line 452: Could be also `debug_assert!()` the reverse lexicographic ordering?

cospectrum: Ok, one moment

cospectrum: It would be cool if `is_sorted` was in stable

Morgane55440: what about
```
elements.iter().tuple_windows().all(|(p1, p2)| (p1.y == p2.y && p1.x < p2.x) || p1.y < p2.y)

```
it checks both for unity and order with O(1) space

theotherphil:
Thanks!

Issue #237, opened 2017-10-15
Gamma correctness


theotherphil:
The code in this library assumes that the input intensities are encoded in a linear colour space, but in the majority of cases they won't be. What's the best way to handle this? In many cases the user might not care about the distinction, but we should arrange the types/traits so that you can't accidentally ignore the fact that you're treating non-linear inputs as if they were linear.

Reisz:
Gamma correctness is actually more important, than most people think ([see here](http://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/)).

In my opinion, every image processing API should work in linear space (as the algorithms expect) and convert from/to gamma-encoded on loading/storing. I would suggest the following API for loading/storing:
- `load(path)` Loads image and decodes from sRGB
- `load_enc(path, encoding: Encoding)` Loads image and decodes using various encodings. Options should include linear, as many non-color texture maps in 3d rendering are stored in linear space.

And the same for `store` of course.

Encoding example:
```rust
enum Encoding {
  Linear , sRGB, PlainGamma(f32)
}
```

theotherphil:
I agree that it's important to get this correct. I don't have time to look at this myself at the moment. I suspect there's a bit of thinking to be done about how best to handle linear colour spaces - there's a lot of code in this crate that assumes that pixels have eight bits per channel, and just converting everything into 8bpp linear space and back might result in unacceptable loss of fidelity in the most distinguishable intensities. I'm open to suggestions (or PRs!)

PR #606, opened 2024-05-15
Color Bilateral Filter

ripytide:
This PR expands the implementation of `bilateral_filter()` to any pixel type including `Rgb` by using a `ColorSimilarity` trait which allows arbitrary user defined metrics to be used. Another trait `WeightedAverage` was needed in order to allow the implementation to work with any pixel type.

It's mildly slower than the old version on greyscale images from my testing which was annoying but I can barely even understand the old version as the variable naming was very poor. I think this kind of function can be easily parallelized though which would make it tons faster than the old version.


ripytide:
Here is an example of it in action:
Before:
leaf
After:
bilateral_color

theotherphil:
Thanks, this is a nice function to have. I’ll review it this weekend.

The old function looks to have a very similar structure to yours except for quantising colour differences to allow the use of lookup tables for both colour and distance weights.

ripytide:
Using a lookup table for non-greyscale becomes a much trickier as there are many many different combinations. Also the old version only allowed Gaussian Euclidean similarity but this version allows user-specified Similarity which also makes creating a lookup table more difficult.

src/filter/bilateral.rs
```diff
@@ -0,0 +1,303 @@
...
+        let euclidean_distance = if P::CHANNEL_COUNT == 1 {
+            f32::from(pixel1.channels()[0]) - f32::from(pixel2.channels()[0]).abs()
+        } else {
+            pixel1
+                .channels()
+                .iter()
+                .zip(pixel2.channels().iter())
+                .map(|(c1, c2)| (f32::from(*c1) - f32::from(*c2)).powi(2))
+                .sum::<f32>()
+                .sqrt()
```
theotherphil, line 95: This sqrt is immediately undone by the squaring within `gaussian_weight` - we should be able to get a bit of performance improvement by avoiding this unnecessary work.

src/filter/bilateral.rs
```diff
@@ -0,0 +1,303 @@
...
+    let radius = i16::from(radius);
+
+    let window_range = -radius..=radius;
+    let spatial_similarity_lookup = window_range
+        .clone()
+        .cartesian_product(window_range.clone())
+        .map(|(w_y, w_x)| {
+            //The gaussian of the euclidean spatial distance
+            gaussian_weight(
+                (f32::from(w_x).powi(2) + f32::from(w_y).powi(2)).sqrt(),
```
theotherphil, line 168: Same comment as for `color_similarity` above.

src/filter/mod.rs
```diff
@@ -276,9 +156,11 @@ impl<'a, K: Num + Copy + 'a> Kernel<'a, K> {
     }
 }
 
+/// The gaussian probability density function with a mean of zero.
 #[inline]
-fn gaussian(x: f32, r: f32) -> f32 {
-    ((2.0 * f32::consts::PI).sqrt() * r).recip() * (-x.powi(2) / (2.0 * r.powi(2))).exp()
+fn gaussian_pdf(x: f32, sigma: f32) -> f32 {
+    // Equation derived from <https://en.wikipedia.org/wiki/Gaussian_function>
+    (sigma * f32::consts::TAU.sqrt()).recip() * (-x.powi(2) / (2.0 * sigma.powi(2))).exp()
```
theotherphil, line 163: Let's leave this as `PI`. `TAU` is far more obscure.

tests/regression.rs
```diff
@@ -792,19 +792,6 @@ fn test_hough_line_detection() {
...
-fn test_bilateral_filter() {
-    fn filter(image: &GrayImage) -> GrayImage {
-        let sigma_color: f32 = 20.;
-        let sigma_spatial: f32 = 2.;
-        let radius: f32 = 3. * sigma_spatial;
-        let win_size = (radius * 2. + 1.) as u32;
-        bilateral_filter(image, win_size, sigma_color, sigma_spatial)
-    }
-
-    compare_to_truth_with_tolerance("lumaphant.png", "lumaphant_bilateral.png", filter, 1)
```
theotherphil, line 805: I'd like a regression test for this function.

Running `REGENERATE=1 cargo test` will regenerate the truth images.

src/filter/median.rs
```diff
@@ -323,15 +323,8 @@ struct HistSet {
...
-        // Can't use vec![[0u32; 256], num_channels as usize]
-        // because arrays of length > 32 aren't cloneable.
-        let mut data = Vec::with_capacity(num_channels as usize);
-        for _ in 0..num_channels {
-            data.push([0u32; 256]);
-        }
-
         HistSet {
-            data,
+            data: vec![[0u32; 256]; num_channels.into()],
```
theotherphil, line 327: It would be slightly nicer to merge this change and the change to stats.rs as standalone PRs rather than as part of this one. But this is only a very mild preference - I'm happy to merge them here.

src/filter/bilateral.rs
```diff
@@ -0,0 +1,303 @@
...
+) -> Image<P>
+where
+    I: GenericImage<Pixel = P>,
+    P: WeightedAverage + image::Pixel,
+    C: ColorSimilarity<P>,
+{
+    assert!(!image.width() > i32::MAX as u32);
+    assert!(!image.height() > i32::MAX as u32);
+    assert_ne!(image.width(), 0);
+    assert_ne!(image.height(), 0);
```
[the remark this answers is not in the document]
ripytide: I was hoping `GenericImageView` or a similar trait would have an `is_empty()` default implementation but alas not. We might have to make one ourselves

src/filter/bilateral.rs
```diff
@@ -0,0 +1,303 @@
...
+/// Un-normalized Gaussian Weight
+fn gaussian_weight(x: f32, sigma: f32) -> f32 {
+    (-0.5 * x.powi(2) / sigma.powi(2)).exp()
+}
+
+/// A trait which provides a similarity metric between two pixels based on their colors.
+///
+/// This trait is used with the `bilateral_filter()` function.
+pub trait ColorSimilarity<P> {
+    /// Returns a measure of similarity between the two pixels based on their colors
```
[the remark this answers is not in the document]
ripytide: Great catch, I like the word "similarity" more than "distance" but it also reverses the directionality which I forgot about. I guess I'll revert back to `ColorDistance`.

src/filter/bilateral.rs
```diff
@@ -0,0 +1,303 @@
...
+#[must_use = "the function does not modify the original image"]
+pub fn bilateral_filter<I, P, C>(
+    image: &I,
+    radius: u8,
+    spatial_sigma: f32,
+    color_similarity: C,
+) -> Image<P>
+where
+    I: GenericImage<Pixel = P>,
+    P: WeightedAverage + image::Pixel,
```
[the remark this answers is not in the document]
ripytide: True, I'm wondering about different color-spaces that don't have linear relationships between colors and so have different interpolations between colors that aren't linear. But then this trait might have to be user-defined which is isn't currently so I'm a bit split on this.

src/filter/mod.rs
ripytide: My thoughts were purely on code performance and not readability, but the compiler probably optimizes it anyway so I'll revert to `2*pi`.

tests/regression.rs
ripytide: Ah ok the new implementation does have so pixel rounding differences so I was worried about exact compatibility but I didn't realize I could regenerate the ground truths.

src/filter/median.rs
ripytide: True, it was more laziness on my part as I was mid-PR when I found the change. I'll make a mini-PR for it.

src/filter/bilateral.rs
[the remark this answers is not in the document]
theotherphil: This is another topic with a long-neglected issue: https://github.com/image-rs/imageproc/issues/237

A large fraction of the functions in this library will do the wrong thing if you're not working in linear colour space. (And most of the examples apply them to sRGB image.... oops.)

I think the best approach here is just to make it clear at the crate level that your images need to be in linear space for functions to do the right thing, and include clear docs that explain 1. that most of the images will have to hand are going to be encoded in something close to sRGB, and 2. how to convert from that to linear RGB. I don't think it's worth the complexity of trying to deal with this distinction at the type level - users who care about correctness can easily make sure they're working in linear space and users who don't care shouldn't have to jump through any hoops.

src/filter/mod.rs
theotherphil: 👍 This should be the easiest possible case for constant folding so I expect both versions to produce identical code.

src/filter/bilateral.rs
ripytide: Good catch!

src/filter/bilateral.rs
ripytide: I've switched back to distance here too

ripytide:
All comments raised have been addressed and #620 opened for the non-related changes.

src/filter/bilateral.rs
```diff
@@ -0,0 +1,288 @@
...
+    *P::from_slice(
+        &channel_averages
+            .map(<f32 as AsPrimitive<P::Subpixel>>::as_)
+            .collect_vec(),
+    )
+}
+
+/// Un-normalized Gaussian Weight
+fn gaussian_weight(x: f32, sigma_squared: f32) -> f32 {
+    (-0.5 * x.powi(2) / sigma_squared).exp()
```
theotherphil, line 191: First argument is now actually x_squared so shouldn’t be squared again here

src/filter/bilateral.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_bilateral_filter_greyscale() {
```
theotherphil, line 199: Let’s leave this test input and expectations the same as it was before as a sanity check - a new implementation shouldn’t change the output values enough to make this fail

theotherphil (commented):
This looks almost good to go - just one bug I’ve spotted.

src/filter/bilateral.rs
ripytide: I've left the input image and input parameters the same or as their equivalents since switching from `window_size` to `radius`, hence 3 -> 1, but I think rounding differences in the new implementation are causing the output to change slightly. From my testing it still looks to be giving the same output as shown in the images I posted so I'm fairly confident it's doing the right thing.

theotherphil:
Thanks!

PR #610, opened 2024-05-19
Add docs.rs feature description

sunsided:
This adds the `#[cfg_attr(docsrs, doc(cfg(feature = "...")))]` markers to inform docs.rs about feature conditions. It looks somewhat like this;

image

Since this is a feature that only works with docs.rs, it's gated by the `#![cfg_attr(docsrs, feature(doc_cfg))]` attribute in `lib.rs`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,7 @@
 //! [image]: https://github.com/image-rs/image
 #![deny(missing_docs)]
 #![cfg_attr(test, feature(test))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 #![allow(
     clippy::cast_lossless,
     clippy::too_many_arguments,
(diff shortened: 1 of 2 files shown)


theotherphil:
This is great, thanks - I wasn't aware of this feature.

Before I merge this PR please can you either add this attribute to the rest of the feature-enabled functions in this crate or create a GitHub issue to track that someone needs to do so.

sunsided:
Can do. I didn't add it to the ones that were test-related, but I guess it doesn't harm either.

theotherphil:
Ah, the other uses I was thinking of are actually hidden inside functions, e.g. https://github.com/image-rs/imageproc/blob/44bbf5e7efd8d9de41ffafc2be2a0696cfecfb4b/src/contrast.rs#L245

Maybe there are no missed uses then!

sunsided:
Hum, I think my IDE tricked me and there are indeed some missing. I'll push to the branch again and re-open the PR then. ✌🏼

fintelia:
You can also do `#![cfg_attr(docsrs, feature(doc_auto_cfg))]` to automatically apply the annotation everywhere

theotherphil:
Thanks @fintelia - definitely worth a look

PR #611, opened 2024-05-19
Gradient Refactor V3

ripytide:
I think I made a mistake in #601 by changing too much at once. This time I've focused on some of the core changes and discarded some others (such as no kernel trait this time).

Summary:
- Moved associated `filter()` method to free-standing `filter()` function
- Added a `filter_clamped()` function which was what `filter3x3()` was doing previously
- Removed `filter3x3()` since `filter_clamped()` is identical and works with any dimensions
- Moved all kernel constants to the `kernel` module as associated constants of `Kernel` rather than free-standing constants
- Removed `sobel_vertical()` and similar functions as they are simply applications of `filter_clamped()` with different kernels which is not very maintainable as the number of different kernels grows.
- Added `FOUR_LAPLACIAN_3X3` and `EIGHT_LAPLACIAN_3X3` kernel constants.

### DIFF
--- a/src/kernel.rs
+++ b/src/kernel.rs
@@ -95,7 +95,6 @@ pub const FOUR_LAPLACIAN_3X3: Kernel<'static, i16> = Kernel {
     width: 3,
     height: 3,
 };
-
 /// The 8-connectivity laplacian 3x3 kernel.
 pub const EIGHT_LAPLACIAN_3X3: Kernel<'static, i16> = Kernel {
     data: &[1, 1, 1, 1, -8, 1, 1, 1, 1],
--- a/src/kernel.rs
+++ b/src/kernel.rs
@@ -27,11 +27,17 @@ impl<'a, K> Kernel<'a, K> {
     }
 
     /// Get the value in the kernel at the given `x` and `y` position.
+    ///
+    /// # Panics
+    ///
+    /// If the `x` or `y` is outside of the width or height of the kernel.
     #[inline]
-    pub fn get_unchecked(&self, x: u32, y: u32) -> &K {
+    pub fn at(&self, x: u32, y: u32) -> &K {
         &self.data[(y * self.width + x) as usize]
     }
+}
 
+impl<'a> Kernel<'a, i32> {
     /// The sobel horizontal 3x3 kernel.
     pub const SOBEL_HORIZONTAL_3X3: Kernel<'static, i32> = Kernel {
         data: &[-1, 0, 1, -2, 0, 2, -1, 0, 1],
--- a/src/kernel.rs
+++ b/src/kernel.rs
@@ -37,70 +37,68 @@ impl<'a, K> Kernel<'a, K> {
     }
 }
 
-impl<'a> Kernel<'a, i32> {
-    /// The sobel horizontal 3x3 kernel.
-    pub const SOBEL_HORIZONTAL_3X3: Kernel<'static, i32> = Kernel {
-        data: &[-1, 0, 1, -2, 0, 2, -1, 0, 1],
-        width: 3,
-        height: 3,
-    };
-    /// The sobel vertical 3x3 kernel.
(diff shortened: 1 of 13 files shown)


src/kernel.rs
```diff
@@ -0,0 +1,100 @@
...
+        Kernel {
+            data,
+            width,
+            height,
+        }
+    }
+
+    /// Get the value in the kernel at the given `x` and `y` position.
+    #[inline]
+    pub fn get_unchecked(&self, x: u32, y: u32) -> &K {
```
theotherphil, line 31: `get_unchecked` is typically used for unsafe indexing. I think this is just `get`.

src/kernel.rs
```diff
@@ -0,0 +1,100 @@
...
+        &self.data[(y * self.width + x) as usize]
+    }
+
+    /// The sobel horizontal 3x3 kernel.
+    pub const SOBEL_HORIZONTAL_3X3: Kernel<'static, i32> = Kernel {
```
theotherphil, line 36: These should be top level constants. Being able to refer to a kernel of `i32` values via e.g. `Kernel::<u64>::SOBEL_HORIZONTAL_3x3` is very misleading.

theotherphil:
This broadly looks good to me.

The two main changes I'd like are:
* To move the kernel constant (Sobel, etc.) back up to module level.
* Reinstate the deleted functions except for `Kernel::filter`, and optionally mark them as deprecated - I'd a bit wary of causing user churn for the sake of a few one-line functions.

src/kernel.rs
ripytide: I'm a bit confused why the compiler even requires the disambiguation at all

ripytide: Actually what about implementing as associated constants of Kernel<i32> instead of impl<K> Kernel<K>?

src/kernel.rs
ripytide: To play devil's advocate the theoretical `get_checked` would return an option in case the given indices were out of range rather than panicking which mirrors the various checked Vs unchecked functions in the standard library.

ripytide: To play devil's advocate the theoretical `get_checked` would return an option in case the given indices were out of range rather than panicking which mirrors the various checked Vs unchecked functions in the standard library.

theotherphil: In the standard library `get_unchecked` indicates unchecked indexing rather than indexing that might panic, e.g.

https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get_unchecked
https://doc.rust-lang.org/std/primitive.str.html#method.get_unchecked

Good point that `get` typically returns an `Option`. Maybe this could just be called `at`?

src/kernel.rs
theotherphil: `kernel::Kernel::<i32>::SOBEL_HORIZONTAL_3x3` seems worse than `kernel::SOBEL_HORIZONTAL_3x3`. I don't see what the benefit is from nesting the consts.

ripytide: Nesting is necessary if for if we ever require kernels of different types other than `i32`. It's that or add types to the name of the constants such as `kernel::SOBEL_HORIZONTAL_3X3_i16` which seems worse.

ripytide: However, if you're confident that we will never need non-`i32` kernels or that this change should be postponed until we do need them then I'd be happy to make the constants module-level.

theotherphil: Let's go with module level rather than trying to future proof this. Thanks.

ripytide:
I've implemented the requested changes, are the deprecation warnings ok?

theotherphil:
Looks good to me, thanks!

Just needs the version updating to 0.26 in Cargo.toml to get the semver check to pass.

theotherphil:
Third time lucky! Thanks - I think exposing `filter` and `filter_clamped` this way is a definite improvement.

PR #619, opened 2024-05-19
Re-enable zero-prefixed-literals lint

theotherphil:


theotherphil:
Removing these is a regression in test readability so I'll leave the lint disabled after all

PR #624, opened 2024-05-20
Reduce code duplication in horizontal and vertical filter functions

sunsided:
Following up on #608, I took the liberty of reducing some code duplication in the `horizontal_filter` and `vertical_filter` functions. 

Do note that this branches off #608 itself and therefore looks bigger than it actually is. The _actual_ change is the introduction of the `clamp_and_reset` function. [This comparison] gives a better idea.

[This comparison]: https://github.com/image-rs/imageproc/compare/2b8fb4159ec1d1fc0e5eb6172896080d239ee078..893c8869c0ef06135d349bc412be04ac8b40eb73

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -289,7 +289,7 @@ where
     P: WithChannel<S>,
     K: Num + Copy + Send + Sync + From<P::Subpixel>,
 {
-    filter_parallel(image, kernel, |x| S::clamp(x))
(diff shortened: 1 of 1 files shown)


theotherphil:
This looks good to merge after it’s been rebased.

sunsided:
@theotherphil Done. 🫡

theotherphil:
Thanks!

PR #627, opened 2024-05-20
Fix Cargo.toml `rayon` feature and Update Dependencies

ripytide:
The way the features were defined meant that using `imageproc` with `default-features = false, features = ["rayon"]` would fail to compile as that wouldn't force `image` to use its `rayon` feature. This PR fixes that and updates the cargo dependencies while at it.


theotherphil:
Thanks!

Issue #622, opened 2024-05-20
Remove non-`_mut` function variants.


ripytide:
A lot of functions at the moment have two variants: a `_mut` version and a non-`_mut` version.
For example, `morphology::dilate()` and `morphology::dilate_mut()`. The non-`_mut` version is simply a 3-line wrapper of the `mut` version.

This makes maintaining the library a lot more difficult as it doubles the API surface for those functions. Documenting both functions becomes more difficult as you either have to duplicate the docs but just change from `&mut image` to passing in a whole image and storing the result, or, link one of the docs to the others using a doc-link and hope the user can figure out the differences in ownership semantics.

This seems like the kind of job a user should be taking care of and not this library. If the user is doing `&mut image.clone()` a whole lot then they can easily abstract that themselves but I don't feel we should be taking on that burden for them.

My suggestion would be to either deprecate all the non-`mut` function variants and then delete them in the second-next release or simply delete them outright.

ripytide:
This came up when working on #602 where the non-`mut` functions double the amount of work I would have to do.

ripytide:
Judging by some [grep.app](https://grep.app/search?q=imageproc%3A%3A&filter[lang][0]=Rust&filter[path.pattern][0]=src) searches it varies quite a bit.

[dilate](https://grep.app/search?q=morphology%3A%3Adilate&filter[lang][0]=Rust&filter[path.pattern][0]=src) shows only the mut version is used.
[map_pixels](https://grep.app/search?q=imageproc%3A%3Amap&filter[lang][0]=Rust&filter[path.pattern][0]=src) shows both the mut version and the non-mut version.

ripytide:
> In theory the generation of the _mut versions could be automated via a proc macro but this feels unlikely to be worth the extra complexity or compile time.

I think a proc macro would solve my issues with maintainability and help with standardization, I don't think the added compile time would be noticeable for such a simple proc-macro, the complexity would be mostly a one-time cost on the implementation of the proc-macro and should be very easy to use after its created?

ripytide:
Nevermind, I tried implementing the proc-macro and I needed to use a lot of heuristics when generating the non-mut function signature such as choosing a return type based on the input image type which could get very complicated very quickly. I'll close this issue then and we'll keep the status quo.

What about switching the documentation to the mut version rather than the non-mut function seeing as that is the "real" function whereas the non-mut variation is simply a useful helper function? Then we could standardize the documentation of the non-mut functions as:

```rust
[...]

PR #632, opened 2024-05-21
Add a doc-macro for mut versions of functions.

ripytide:
As discussed in #622.

As it turned out the `imageproc::map` functions were the only functions not able to be de-duplicated using a doc-comment generating macro as the image bounds are completely different per function so complete separate implementations (`Fn(P) -> P` vs `Fn(P) -> Q).

But overall I'm very happy with the amount of de-duplication achieved which should make things much more standardized and easier to maintain and extend.

I also standardized the non-mut functions being above the mut function in the source code with no empty-line between them for consistency.


theotherphil:
Thanks, this is a nice simplification.

PR #635, opened 2024-05-22
implement the parallel doc macro

### LINKED ISSUE #634: Another doc-macro for parallel variations of functions?
Should we create another doc-comment generating macro (as in #632) for the `_parallel()` function versions? This would have the same benefits of easier maintenance and better consistency.

ripytide:
Closes #634

### DIFF
--- a/src/doc_macros.rs
+++ b/src/doc_macros.rs
@@ -19,7 +19,7 @@ macro_rules! generate_mut_doc_comment {
 }
 
 /// A macro for generating the doc-comments for parallel versions of various
-/// image processing functions. It takes the name of then non-parallel function as an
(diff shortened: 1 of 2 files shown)


src/doc_macros.rs
```diff
@@ -17,3 +17,21 @@ macro_rules! generate_mut_doc_comment {
         )
     };
 }
+
+/// A macro for generating the doc-comments for parallel versions of various
+/// image processing functions. It takes the name of then non-parallel function as an
```
theotherphil, line 22:
```suggestion
/// image processing functions. It takes the name of the non-parallel function as an
```

theotherphil:
One typo,  but otherwise looks good - thanks!

PR #636, opened 2024-05-22
Remove noisy `ImageBuffer` type

ripytide:
This PR replaces the `ImageBuffer` type with more relevant `Image` and `GrayImage` types as appropriate.

A lot of the time functions return `Image<P>` anyway so using `Image` inside the function makes more sense to me. Also a lot of the time `ImageBuffer<P, Vec<P>>` is used explicitly which is repetitive and noisy compared to simply `Image<P>`.

The only exception to this is doc-examples which I avoided since it may teach users to rely on the `imageproc::definitions::Image` type synonym which I wasn't sure about. However, I'd be happy to also switch these to use `Image` as well if you think that'd be ok.

### DIFF
--- a/examples/blend.rs
+++ b/examples/blend.rs
@@ -1,7 +1,7 @@
 //! Demonstrates the current incorrect handling of gamma for RGB images
 
-use image::{ImageBuffer, Rgb};
-use imageproc::pixelops::interpolate;
+use image::Rgb;
+use imageproc::{definitions::Image, pixelops::interpolate};
 
 fn main() {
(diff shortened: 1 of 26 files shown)


theotherphil:
I’m a bit ambivalent about this one - it’s nice to save the typing and having to constantly specify the Container type parameter everywhere, but it’s a shame to hide the true type name and require developers to go and learn that this is actually ImageBuffer after all.

Given that we already have the type alias I guess it’s better to use it consistently than inconsistently. (Although I’m tempted to just delete the alias entirely at some point.)

Issue #532, opened 2023-06-15
adaptive_threshold lacks the delta parameter


stephanemagnenat:
The OpenCV equivalent of [`adaptive_threshold`](https://docs.rs/imageproc/latest/imageproc/contrast/fn.adaptive_threshold.html) has delta parameter ([they call it C in the documentation](https://docs.opencv.org/4.x/d7/d1b/group__imgproc__misc.html#ga72b913f352e4a1b1b397736707afcde3)), which is necessary in many use cases.

Currently the imageproc equivalent miss it. I need it, so I rewrote the function. I'll be happy to contribute the extended version in a PR, maybe we can name it `adaptive_threshold_with_delta` and let the old one use the new version with 0 as delta. That way, the API is not broken.

What do you think?

theotherphil:
This would a welcome addition if you’re still up for creating the PR. Thanks.

theotherphil:
Added in https://github.com/image-rs/imageproc/pull/637

PR #637, opened 2024-05-22
Add `delta` parameter to `adaptive_threshold`

Dantsz:
For #532 
The opencv implementation seems to take the delta as float and rounds it differently based on if the thresholding is inverted or not, we don't have an inverted mode for this so passing delta as `i32` should be fine.

### DIFF
--- a/src/contrast.rs
+++ b/src/contrast.rs
@@ -13,7 +13,7 @@ use crate::stats::{cumulative_histogram, histogram};
 
 /// Applies an adaptive threshold to an image.
 ///
-/// Equivalent to `adaptive_threshold_with_delta` with delta=0.
+/// Equivalent to [`adaptive_threshold_with_delta()`] with delta=0.
 pub fn adaptive_threshold(image: &GrayImage, block_radius: u32) -> GrayImage {
     adaptive_threshold_with_delta(image, block_radius, 0)
 }
--- a/src/contrast.rs
+++ b/src/contrast.rs
@@ -463,6 +463,23 @@ mod tests {
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


src/contrast.rs
```diff
@@ -17,6 +17,19 @@ use crate::stats::{cumulative_histogram, histogram};
 /// in the (2 * `block_radius` + 1) square block centered on it. If the pixel is at least as bright
 /// as the threshold then it will have a value of 255 in the output image, otherwise 0.
```
ripytide, line 18: Perhaps these docs could now just say its equivalent to `adaptive_threshold_with_delta()` with a delta=0 to prevent duplicated docs?

src/contrast.rs
```diff
@@ -13,10 +13,21 @@ use crate::stats::{cumulative_histogram, histogram};
 
 /// Applies an adaptive threshold to an image.
 ///
+/// Equivalent to `adaptive_threshold_with_delta` with delta=0.
```
ripytide, line 16:
```suggestion
/// Equivalent to [`adaptive_threshold_with_delta()`] with delta=0.
```

theotherphil:
Thanks! I think it would be better to just add the delta parameter to the existing function - it’s a breaking change but a trivial one for users to deal with.

theotherphil:
We run nightly clippy and it looks like you’ve fallen foul of a newly added lint triggering on old code. However, the fix it recommends is trivial so please can you make the edit suggested and then I’ll merge this PR. Thanks for the PR :-)

theotherphil:
Thanks!

PR #642, opened 2024-05-25
Add `filter_map_parallel`

ripytide:
I've re-based the #608 pr since it was ready for merging and don't want it to be left behind.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -663,7 +663,7 @@ mod tests {
             #[test]
             fn $test_name() {
                 // I think the interesting edge cases here are determined entirely
-                // by the relative sizes of the kernel and the image side length, soz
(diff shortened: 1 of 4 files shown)


src/filter/mod.rs
```diff
@@ -578,7 +663,7 @@ mod tests {
             #[test]
             fn $test_name() {
                 // I think the interesting edge cases here are determined entirely
-                // by the relative sizes of the kernel and the image side length, so
+                // by the relative sizes of the kernel and the image side length, soz
```
theotherphil, line 666:
```suggestion
                // by the relative sizes of the kernel and the image side length, so
```

sunsided:
Thanks!

theotherphil:
Thanks @sunsided and @ripytide

PR #647, opened 2024-05-26
Switch to `image::math::Rect`

### LINKED ISSUE #645: Switch `imageproc::Rect` to `image::Rect`
Related to #638

ripytide:
Fixes #645 

I moved all the current inherent methods to a `RectExt` helper trait since the upstream `Rect` doesn't have those methods, yet.

### DIFF
--- a/examples/drawing.rs
+++ b/examples/drawing.rs
@@ -44,18 +44,14 @@ fn main() {
     draw_line_segment_mut(&mut image, (20f32, 180f32), (20f32, 220f32), white);
 
     // Draw a hollow rect within bounds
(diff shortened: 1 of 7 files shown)


theotherphil:
Thanks for the PR.

All of the `draw_x` functions accept inputs that may be extend outside the image and then only draw the points that are within the image bounds. This is a bit odd if you think of the image as a finite canvas but makes more sense for the original use case for several of these functions which was to annotate photos/video frames with areas of interest identified using these images.

I’d rather keep the consistency between these functions so won’t merge this PR.

PR #648, opened 2024-05-26
Make `Kernel::new` constant

cospectrum:
Users may also need a static `Kernel`!

### DIFF
--- a/src/kernel.rs
+++ b/src/kernel.rs
@@ -10,15 +10,14 @@ pub struct Kernel<'a, K> {
 
 impl<'a, K> Kernel<'a, K> {
     /// Construct a kernel from a slice and its dimensions. The input slice is
-    /// in row-major form.
-    pub fn new(data: &'a [K], width: u32, height: u32) -> Kernel<'a, K> {
+    /// in row-major order.
+    ///
+    /// # Panics
+    /// 1. If `width == 0 || height == 0`.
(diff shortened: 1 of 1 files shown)


theotherphil:
It’s a shame to lose the useful panic message but not the end of the world - I can’t imagine it’s likely that anyone will ever be constructing a Kernel in ways that make it hard to work out the dimensions required.

src/kernel.rs
```diff
@@ -38,66 +37,34 @@ impl<'a, K> Kernel<'a, K> {
...
+pub const ROBERTS_VERTICAL_2X2: Kernel<'static, i32> = Kernel::new(&[0, 1, -1, -0], 2, 2);
 
 /// The 4-connectivity laplacian 3x3 kernel.
-pub const FOUR_LAPLACIAN_3X3: Kernel<'static, i16> = Kernel {
-    data: &[1, 1, 1, 1, -4, 1, 1, 1, 1],
-    width: 3,
-    height: 3,
-};
+pub const FOUR_LAPLACIAN_3X3: Kernel<'static, i16> =
+    Kernel::new(&[1, 1, 1, 1, -4, 1, 1, 1, 1], 3, 3);
```
ripytide, line 67: I think I typed this one wrong as well, it's supposed to have zeros in the corners, whoops.

Issue #649, opened 2024-05-26
`Laplacian` from `master` is wrong


cospectrum:
Why it's `[1, 1, 1, 1, -4, 1, 1, 1, 1]`, while in the doc it's `[0, 1, 0, 1, -4, 1, 0, 1, 0]`?

cospectrum:
No, I think it's ok. We just have a bunch of laplacians now

theotherphil:
Looks like it was broken accidentally in https://github.com/image-rs/imageproc/pull/611 and there are no tests for `laplacian_filter` so this wasn’t caught.

Let’s add a test alongside fixing the kernel definition.

theotherphil:
The bug was added since the last release so no crate versions are impacted.

theotherphil:
Fixed by https://github.com/image-rs/imageproc/pull/650

PR #650, opened 2024-05-26
Fix lalplacian

cospectrum:
Possible fix of https://github.com/image-rs/imageproc/issues/649 ?


theotherphil:
Can you add a test as well to prevent any future refactoring errors.

theotherphil:
I was imagining a test of the outputs but I guess this works too!

Thanks for the fix. I intended to add a test after merging the initial PR for Laplacian filtering but forgot about it in all the follow up discussion. I’ll make sure all functions have tests pre-merge in future.

Issue #644, opened 2024-05-26
Add hardcoded filter3x3 function to benchmark current generic filter function against


theotherphil:

ripytide:
It seems #646 adds an implementation for 3x3, 5x5, 7x7, this might be closable then. Or is there something I'm missing with this issue as compared to #643?

theotherphil:
This issue is to add an implementation of filter3x3 along the lines of what we had when this library was first written - a raw loop that averages the pixel values at 9 fixed offsets from the current coordinate - to check that we don’t accidentally end up with an implementation that’s slower than that for this use case.

cospectrum:
The filter has become 4 times slower

cospectrum:
malloc for every pixel 😮‍💨

theotherphil:
@cospectrum which versions are you comparing?

cospectrum:
> @cospectrum which versions are you comparing?

`master` and `0.25`

theotherphil:
Ouch. I’ll take a look today. @ripytide has benchmarks showing no impact at https://github.com/image-rs/imageproc/pull/608#issuecomment-2122314598 so something is odd here.

theotherphil:
Comparing v0.25.0 and current master (5e00bd1a6df330f7be76e556bce89c1c1080a40a) on a 2021 MacBook Pro with an M1 Max chip.

The function signatures and set of benchmarks have both changed across this range so I've
* Created a filter5x5 and filter7x7 function in version 0.25 which are just copy-pasted from filter3x3 with a different kernel size
* Added the following benchmarks

```
// For v0.25
    #[bench]
    fn bench_filter3x3_i32_filter(b: &mut Bencher) {
        let image = gray_bench_image(500, 500);
        #[rustfmt::skip]
        let kernel: Vec<i32> = vec![
            -1, 0, 1,
            -2, 0, 2,
            -1, 0, 1
        ];

        b.iter(|| {
            let filtered: ImageBuffer<Luma<i16>, Vec<i16>> =
                filter3x3::<_, _, i16>(&image, &kernel);
            black_box(filtered);
        });
    }

    #[bench]
    fn bench_filter5x5_i32_filter(b: &mut Bencher) {
        let image = gray_bench_image(500, 500);
        #[rustfmt::skip]
        let kernel: Vec<i32> = vec![
            -1, 0, 1, 2, 3,
            -2, 0, 2, 3, 5,
            -1, 0, 1, 6, 9,
[...]

PR #654, opened 2024-05-27
Replace `filter` with the old implementation

cospectrum:
Related https://github.com/image-rs/imageproc/issues/644

### DIFF
--- a/src/kernel.rs
+++ b/src/kernel.rs
@@ -2,13 +2,19 @@
 
 /// An borrowed 2D kernel, used to filter images via convolution.
 #[derive(Debug, Copy, Clone)]
-pub struct Kernel<'a, K> {
+pub struct Kernel<'a, K>
+where
+    K: Copy,
+{
     pub(crate) data: &'a [K],
     pub(crate) width: u32,
     pub(crate) height: u32,
 }
 
-impl<'a, K> Kernel<'a, K> {
+impl<'a, K> Kernel<'a, K>
+where
+    K: Copy,
+{
     /// Construct a kernel from a slice and its dimensions. The input slice is
     /// in row-major order.
     ///
@@ -31,8 +37,10 @@ impl<'a, K> Kernel<'a, K> {
     ///
     /// If the `x` or `y` is outside of the width or height of the kernel.
     #[inline]
-    pub fn at(&self, x: u32, y: u32) -> &K {
-        &self.data[(y * self.width + x) as usize]
+    pub(crate) fn at(&self, x: u32, y: u32) -> K {
+        debug_assert!(x < self.width);
+        debug_assert!(y < self.height);
+        self.data[(y * self.width + x) as usize]
     }
 }
 
--- a/src/kernel.rs
+++ b/src/kernel.rs
@@ -35,12 +35,24 @@ where
     ///
     /// # Panics
     ///
-    /// If the `x` or `y` is outside of the width or height of the kernel.
+    /// If the `y * kernel.width + x` is outside of the kernel data.
     #[inline]
-    pub(crate) fn at(&self, x: u32, y: u32) -> K {
+    pub fn get(&self, x: u32, y: u32) -> K {
         debug_assert!(x < self.width);
         debug_assert!(y < self.height);
-        self.data[(y * self.width + x) as usize]
+        let at = usize::try_from(y * self.width + x).unwrap();
+        self.data[at]
+    }
+
+    /// Get the value in the kernel at the given `x` and `y` position, without
+    /// doing bounds checking.
+    #[inline]
+    pub unsafe fn get_unchecked(&self, x: u32, y: u32) -> K {
+        debug_assert!(x < self.width);
+        debug_assert!(y < self.height);
+        let at = usize::try_from(y * self.width + x).unwrap();
+        debug_assert!(at < self.data.len());
+        self.data.get_unchecked(at).clone()
     }
 }
 
--- a/src/kernel.rs
+++ b/src/kernel.rs
@@ -46,13 +46,16 @@ where
 
     /// Get the value in the kernel at the given `x` and `y` position, without
     /// doing bounds checking.
+    ///
+    /// # Safety
+    /// The caller must ensure that `y * self.width + x` is in bounds of kernel data.
     #[inline]
     pub unsafe fn get_unchecked(&self, x: u32, y: u32) -> K {
         debug_assert!(x < self.width);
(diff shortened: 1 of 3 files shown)


ripytide:
Can you show a benchmark comparison on your machine with a before/after this pr? I'll do the same on my machine as well.

theotherphil:
@ripytide I've just added a benchmark comparison on my machine on https://github.com/image-rs/imageproc/issues/644#issuecomment-2133278304

ripytide:
My Results on a `Ryzen 5 4500U`:
```
 name                                                     before1 ns/iter  after1 ns/iter  diff ns/iter   diff %  speedup 
 filter::benches::bench_box_filter                        2,069,851        2,067,948             -1,903   -0.09%   x 1.00 
 filter::benches::bench_filter_clamped_gray_3x3           1,235,373        1,187,812            -47,561   -3.85%   x 1.04 
 filter::benches::bench_filter_clamped_gray_5x5           2,860,594        2,790,122            -70,472   -2.46%   x 1.03 
 filter::benches::bench_filter_clamped_gray_7x7           4,821,814        5,087,034            265,220    5.50%   x 0.95 
 filter::benches::bench_filter_clamped_parallel_gray_3x3  780,587          459,600             -320,987  -41.12%   x 1.70 
 filter::benches::bench_filter_clamped_parallel_gray_5x5  962,982          927,658              -35,324   -3.67%   x 1.04 
 filter::benches::bench_filter_clamped_parallel_gray_7x7  1,402,118        1,595,752            193,634   13.81%   x 0.88 
 filter::benches::bench_gaussian_f32_stdev_1              286,052          295,040                8,988    3.14%   x 0.97 
 filter::benches::bench_gaussian_f32_stdev_10             1,377,759        1,384,059              6,300    0.46%   x 1.00 
 filter::benches::bench_gaussian_f32_stdev_3              524,585          527,970                3,385    0.65%   x 0.99 
 filter::benches::bench_horizontal_filter                 902,045          913,429               11,384    1.26%   x 0.99 
 filter::benches::bench_separable_filter                  663,439          705,558               42,119    6.35%   x 0.94 
 filter::benches::bench_vertical_filter                   934,897          941,840                6,943    0.74%   x 0.99 
```

theotherphil:
@ripytide can you also try with the benchmarks listed on https://github.com/image-rs/imageproc/issues/644#issuecomment-2133278304 so we can get a consistent comparison between my benchmarking and yours.

ripytide:
So on my machine the only massive change is the parallel_3x3 which is `1.7x` but everything else is quite similar and goes either way. But definitely not `4x`.

cospectrum:
```rs
use std::{hint::black_box, time::Instant};
use imageproc::image::RgbImage;

const TRIES: usize = 20;

fn main() {
    const W: u32 = 2600;
    const H: u32 = W;
    
    let img = black_box(RgbImage::new(W, H));
    let kernel = black_box([0i32; 3 * 3]);

    let t = Instant::now();
    for _ in 0..TRIES {
        let _: RgbImage = black_box(imageproc::filter::filter3x3(&img, &kernel));
    }
    dbg!(t.elapsed());

    let ker = imgproc::kernel::Kernel::new(&kernel, 3, 3);
    let t_new = Instant::now();
    for _ in 0..TRIES {
        let _: RgbImage = black_box(imgproc::filter::filter_clamped(&img, ker));
    }
    dbg!(t_new.elapsed());
}
```

```toml
[dependencies]
imageproc = "0.25.0"
imgproc = { path = "../imageproc/", package = "imageproc" }
```

cospectrum:
```
[src/main.rs:17:5] t.elapsed() = 2.099322542s
[src/main.rs:24:5] t_new.elapsed() = 8.245356375s
```

ripytide:
Interesting, I wonder what the difference between that test and the `3x3` in the benchmarks is that's making such a big difference.

theotherphil:
The regressions are much larger for RGB images: https://github.com/image-rs/imageproc/issues/644#issuecomment-2133356957

Which might explain some of the difference between what your respective benchmarks are showing. I expect we're missing benchmarks on RGB images for quite a lot of functions in this crate!

cospectrum:
@theotherphil By the way, what should happen if `P::CHANNEL_COUNT != Q::CHANNEL_COUNT`?

cospectrum:
The `zip` will be trimmed to the minimum length.

ripytide:
Should probably panic in that scenario, is there a compile-time assert function?

cospectrum:
Compile time assert is `const _: () = assert!(expr)`

cospectrum:
But we can't use this, for the same reason we can't allocate the stack by the number of channels

cospectrum:
The only way to do it in stable Rust is to add `typenum` in `Pixel` trait, I think

theotherphil:
@cospectrum I’d go with a runtime assert to check that channel counts match. Seems like a less likely user error than many of the places we already use panics to check preconditions.

cospectrum:
> @cospectrum I’d go with a runtime assert to check that channel counts match. Seems like a less likely user error than many of the places we already use panics to check preconditions.

Done

theotherphil:
@cospectrum , @ripytide as both the function signatures and the set of benchmarks have changed since 0.25 comparisons across the two versions are a bit of a chore.

So I'll merge this, check manually that performance now matches 0.25, and we can then use this commit as the baseline for future changes.

src/kernel.rs
```diff
@@ -2,13 +2,19 @@
 
 /// An borrowed 2D kernel, used to filter images via convolution.
 #[derive(Debug, Copy, Clone)]
-pub struct Kernel<'a, K> {
+pub struct Kernel<'a, K>
+where
+    K: Copy,
+{
```
ripytide, line 8: Having bounds in struct definitions is bad practice, bounds should appear where they are used.

src/kernel.rs
```diff
@@ -29,10 +35,28 @@ impl<'a, K> Kernel<'a, K> {
     ///
     /// # Panics
     ///
-    /// If the `x` or `y` is outside of the width or height of the kernel.
+    /// If the `y * kernel.width + x` is outside of the kernel data.
     #[inline]
-    pub fn at(&self, x: u32, y: u32) -> &K {
-        &self.data[(y * self.width + x) as usize]
+    pub fn get(&self, x: u32, y: u32) -> K {
```
ripytide, line 40: Why did you rename `at()` to `get()`? This was discussed as not being a good idea as `get()` normally returns an option.

PR #655, opened 2024-05-27
Add rgb benchmarks to for the filter function

ripytide:
Adds `rgb` variants to the `filter_clamped()` function.

This should make benchmarking the full performance effect of implementation changes much easier.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -1034,10 +1034,28 @@ mod benches {
             }
         };
     }
+    macro_rules! bench_filter_clamped_rgb {
+        ($name:ident, $kernel_size:expr, $image_size:expr, $function_name:ident) => {
+            #[bench]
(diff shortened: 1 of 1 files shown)


src/filter/mod.rs
```diff
@@ -1034,10 +1034,28 @@ mod benches {
             }
         };
     }
+    macro_rules! bench_filter_clamped_rgb {
+        ($name:ident, $kernel_size:expr, $image_size:expr, $function_name:ident) => {
+            #[bench]
+            fn $name(b: &mut Bencher) {
+                let image = rgb_bench_image(500, 500);
```
theotherphil, line 1041:
```suggestion
                let image = rgb_bench_image($image_size, $image_size);
```

theotherphil:
Thanks!

PR #656, opened 2024-05-27
Removed allocation from `filter_pixel()`

ripytide:
Achieved via always computing with 4 channels using an array of length 4 since const generic expressions are unstable.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -109,8 +109,6 @@ where
     F: Fn(K) -> Q::Subpixel,
     K: num::Num + Copy + From<P::Subpixel>,
 {
-    //TODO refactor this to use the `crate::maps` functions once that lands
-
     let (width, height) = image.dimensions();
 
     let mut out = Image::<Q>::new(width, height);
@@ -134,8 +132,6 @@ where
     F: Fn(K) -> Q::Subpixel + Send + Sync,
     K: Num + Copy + From<P::Subpixel> + Sync,
 {
-    //TODO refactor this to use the `crate::maps` functions once that lands
-
     use rayon::iter::IndexedParallelIterator;
     use rayon::iter::ParallelIterator;
 
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -23,6 +23,10 @@ use std::cmp::{max, min};
 use std::f32;
 
 /// Calculates the new pixel value for a particular pixel and kernel.
+///
+/// # Panics
+///
+/// If `P::CHANNEL_COUNT != Q::CHANNEL_COUNT`
 fn filter_pixel<P, K, F, Q>(x: u32, y: u32, kernel: Kernel<K>, f: F, image: &Image<P>) -> Q
 where
     P: Pixel,
@@ -30,6 +34,8 @@ where
     F: Fn(K) -> Q::Subpixel,
     K: num::Num + Copy + From<P::Subpixel>,
 {
+    assert_eq!(P::CHANNEL_COUNT, Q::CHANNEL_COUNT);
+
     let (width, height) = (image.width() as i64, image.height() as i64);
     let (k_width, k_height) = (kernel.width as i64, kernel.height as i64);
     let (x, y) = (i64::from(x), i64::from(y));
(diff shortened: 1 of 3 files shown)


ripytide:
Can you push a branch with the code you are running to get that result so I can instigate that?

ripytide:
Okay now that #654 was merged this comparison becomes simpler.
In the generic case (not exactly 3x3) this version reduces the code duplication and is simply faster.
image

ripytide:
@cospectrum can you share your implementation of `filter3x3()` as well so I can try to reproduce your results?

cospectrum:
> @cospectrum can you share your implementation of `filter3x3()` as well so I can try to reproduce your results or are you using the `filter3x3_reference()`?

It's not "my" implementation. It's from 0.25

ripytide:
Okay I think I've tracked down the discrepancy between the `imageproc` benchmarks and your external benchmarks via a separate crate.

I get the same results as you by default. But if you add the same build config as `imageproc` to your external crate I get the same results as `imageproc`.

## Compile Settings
```toml
[profile.release]
opt-level = 3
debug = true

[profile.bench]
opt-level = 3
debug = true
rpath = false
lto = false
debug-assertions = false
codegen-units = 1
```

## Benchmarked Code
```rust
#![feature(test)]

extern crate test;

use std::hint::black_box;

use test::Bencher;

const SIZE: u32 = 300;

fn main() {}

#[bench]
fn bench_025(b: &mut Bencher) {
    let image = black_box(imageproc_025::image::RgbImage::new(SIZE, SIZE));

    let kernel: Vec<i32> = (0..3 * 3).collect();

    b.iter(|| {
        let filtered = imageproc_025::filter::filter3x3::<_, _, i16>(&image, &kernel);
        black_box(filtered);
    });
}
#[bench]
fn bench_master(b: &mut Bencher) {
    let image = black_box(imageproc_master::image::RgbImage::new(SIZE, SIZE));

    let kernel: Vec<i32> = (0..3 * 3).collect();
    let kernel = imageproc_master::kernel::Kernel::new(&kernel, 3, 3);

    b.iter(|| {
        let filtered = imageproc_master::filter::filter_clamped::<_, _, i16>(&image, kernel);
        black_box(filtered);
    });
}
```

# Results

## Master No Settings
```
test bench_025    ... bench:   3,831,394 ns/iter (+/- 10,242)
test bench_master ... bench:   3,902,486 ns/iter (+/- 13,441)
```
## PR No Settings
```
test bench_025    ... bench:   2,179,298 ns/iter (+/- 16,869)
test bench_master ... bench:   7,558,955 ns/iter (+/- 30,916)
```
## Master With Settings
```
test bench_025    ... bench:   2,003,191 ns/iter (+/- 19,839)
test bench_master ... bench:   3,912,089 ns/iter (+/- 8,554)
```
## PR With Settings
```
test bench_025    ... bench:   1,990,703 ns/iter (+/- 10,605)
test bench_master ... bench:   3,253,459 ns/iter (+/- 11,167)
```

ripytide:
Now I suppose I just need to track down which specific build setting is making such a big difference.

ripytide:
When using `debug=false` now the two implementation are both `3,300,000` but `master_025` is still `2,000,000` which doesn't make any sense at all?

cospectrum:
If I bench `0.25` against `master`, `master` will have the same settings as your branch, but `master` has the same speed as `0.25`, while yours will be slower

theotherphil:
I'll try this PR vs the current master on my MacBook when I have time within the next few days.

theotherphil:
@cospectrum what machine are you benchmarking on? I appear to be seeing much larger regressions on an Apple ARM chip than @ripytide is on their Ryzen x86.

cospectrum:
> @cospectrum what machine are you benchmarking on? I appear to be seeing much larger regressions on an Apple ARM chip than @ripytide is on their Ryzen x86.

M1

ripytide:
I think the setting which changes which implementation is faster is `codegen-units`:

Here I've benchmarked the default compile settings vs `codegen-units=1` which is used by the benchmarks normally.
```toml
[profile.release]
codegen-units = 1
```
## Results
```
test master_codegen=1 ... bench:   2,375,138 ns/iter (+/- 3,180)
test pr_codegen=1 ... bench:   1,974,065 ns/iter (+/- 6,905)
test master_codegen=16 ... bench:   2,489,427 ns/iter (+/- 10,247)
test pr_codegen=16 ... bench:   4,585,778 ns/iter (+/- 27,787)
``

**Comment by theotherphil** [comment]
I’ve not had much time in the last couple of weeks but I still intend to benchmark this branch against master on my MacBook.

**Comment by theotherphil** [comment]
```
git checkout master
git pull
cargo +nightly bench filter_clamped_rgb >> bench_master.txt

git fetch origin pull/656/head:pr656
git checkout pr656
cargo +nightly bench filter_clamped_rgb >> bench_pr656.txt
```

Results on master:
```
test filter::benches::bench_filter_clamped_rgb_3x3                                    ... bench:   1,190,189.60 ns/iter (+/- 33,991.58)
test filter::benches::bench_filter_clamped_rgb_5x5                                    ... bench:   2,903,974.95 ns/iter (+/- 10,175.80)
test filter::benches::bench_filter_clamped_rgb_7x7                                    ... bench:   5,348,075.00 ns/iter (+/- 38,132.12)
```

Results for this PR:
```
test filter::benches::bench_filter_clamped_rgb_3x3                                    ... bench:   1,762,945.85 ns/iter (+/- 57,415.80)
test filter::benches::bench_filter_clamped_rgb_5x5                                    ... bench:   3,940,433.40 ns/iter (+/- 10,009.45)
test filter::benches::bench_filter_clamped_rgb_7x7                                    ... bench:   7,041,208.30 ns/iter (+/- 22,404.16)
```

This still has large regressions, at least on an Apple ARM chip, so I'll close this PR.

I'll need to look at filter performance a bit anyway before the next release of this crate, to resolve https://github.com/image-rs/imageproc/issues/664.

PR #658, opened 2024-05-27
Add reference filter3x3

theotherphil:
```
test filter::benches::bench_filter_clamped_gray_3x3                                   ... bench:     704,325.00 ns/iter (+/- 27,325.96)
test filter::benches::bench_filter_clamped_gray_3x3_ref                               ... bench:     230,120.83 ns/iter (+/- 4,152.60)
```

The current Implementation of `filter_clamped` is 3x slower than a simple hardcoded implementation for a grayscale image and 3x3 kernel.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -1025,7 +1025,7 @@ mod benches {
         for y in 0..height {
             let y_prev = std::cmp::max(1, y) - 1;
             let y_next = std::cmp::min(height - 2, y) + 1;
-   
+
             for x in 0..width {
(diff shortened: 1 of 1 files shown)


ripytide:
I bet the compiler is able to do some fancy optimizations when the kernel size is hard-coded as 3. We could have an if statement in the generic `filter()` where if `kernel.width() == 3` and `kernel.height() == 3` then we call `filter3x3()` otherwise we just use the generic version. Then we don't need to expose two different kernel function to the user.

theotherphil:
Yes, we can hide some some manual specialisation behind `filter` as necessary.

PR #660, opened 2024-05-27
Refactor `gradients`

cospectrum:
@theotherphil Did you approve `deprectaed` macro? I don't like it. And `gradients_grayscale` hard to use because there are 3 unused generics

### DIFF
--- a/src/gradients.rs
+++ b/src/gradients.rs
@@ -7,7 +7,7 @@ use crate::map::{ChannelMap, WithChannel};
 use image::{GenericImage, GenericImageView, GrayImage, Luma, Pixel};
 use itertools::multizip;
 
-/// A special version of `gradient` function for grayscale images which doesn't require giving a
+/// A special version of [`gradients`] function for grayscale images which doesn't require giving a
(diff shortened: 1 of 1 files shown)


cospectrum:
> functions are inflexible and pollute the docs

And now they will pollute the user's codebase with long upper case variables and reimplementations

ripytide:
I think it's more readable as well tbh. Easier than having to look up the docs of one of 20 weirdly named functions every time I encounter one. I only have to learn one function `filter_clamped` and then I as a reader will understand every call to it.

theotherphil:
I’ve created https://github.com/image-rs/imageproc/issues/661 to discuss what API we actually want for filtering. So we can agree this in one place rather than going back and forth in pull requests. (I’m also tempted to name `filter_clamped` just `filter` as it seems the more useful function so deserves a shorter name.) I don’t want to keep churning function names on every release as this will cause repeated breaking changes for users and there are for more users than maintainers!

@cospectrum @ripytide I’ll merge this PR but will get consensus on the linked issue before releasing the next crate version.

PR #662, opened 2024-05-30
Add compose module and functions

ripytide:
This PR relies on #647 for the `Rect` type.
 
I have more functions that fit in this category I'd like to add in the future but I thought I'd start with some basic ones.

I've added three main functions `crop()`, `flip_horizontal()` and `flip_vertical()`.

This is part of the wider effort of migrating function from `image::imageops` https://github.com/image-rs/image/issues/2238

### DIFF
--- a/src/compose.rs
+++ b/src/compose.rs
@@ -44,6 +44,7 @@ where
         *image.get_pixel(rect.x + x, rect.y + y)
     })
 }
+#[cfg(feature = "rayon")]
 #[doc = generate_parallel_doc_comment!("crop")]
 pub fn crop_parallel<P>(image: &Image<P>, rect: Rect) -> Image<P>
 where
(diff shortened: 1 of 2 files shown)


theotherphil:
This looks good except for the Rect changes, thanks.

None of these functions are “composing” images in any sense I can think of, but I can’t think of a good name off the top of my head. I’m happy to  merge this without the Rect change and worry about module names before the next version is released

ripytide:
I think I chose the name "compose" from some functions from `imageops`, mainly `replace()` and `overlay()` which combine multiple images hence the name `compose`. Though you're right the mirror functions don't really fit but I couldn't think of any other names at the time so I though I'd just put them all in the `compose` module. I don't think I mind what the modules are called so feel free to change it at any point.

ripytide:
Okay I've removed the `Rect` changes :+1:

theotherphil:
Thanks! It would be nice to have a benchmark for the cropping function to see if the parallelization actually helps, but that isn't a blocker to merging.

Issue #244, opened 2017-10-30
Add regression tests for text drawing functions


theotherphil:
e.g. as described here: https://github.com/PistonDevelopers/imageproc/pull/157#issuecomment-284229911

Make sure they cover the bug fixed here: https://github.com/PistonDevelopers/imageproc/pull/174

PR #663, opened 2024-06-07
improved draw_text_mut performance

Icekey:
I noticed during profiling that most of the time was spend just rounding numbers. It turned out that the min of the bounding box was rounded for every pixel in the "draw_text_mut" function. 

In my testing the fix improved performace by a lot. For example drawing the text in the font.rs example 500000 times went from 22 sec down to 12 sec. The performance improvement is even better for texts with a larger font size. In one of my benchmarks the execution time went from 14.6 sec to 2.7 sec.

I also moved the "gv.clamp(0.0, 1.0)" into the if block, since it was not need before it, but this did not noticably change performance.

### DIFF
--- a/src/drawing/text.rs
+++ b/src/drawing/text.rs
@@ -84,15 +84,17 @@ pub fn draw_text_mut<C>(
     let image_height = canvas.height() as i32;
 
     layout_glyphs(scale, font, text, |g, bb| {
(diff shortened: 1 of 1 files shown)


theotherphil:
Thanks!

(Aside: text drawing regression tests https://github.com/image-rs/imageproc/issues/244 would be really useful to help avoid breakages from refactoring in this area!)

PR #666, opened 2024-06-09
Add `replace()` and `overlay()` functions

ripytide:
Two more functions migrated and re-implemented for https://github.com/image-rs/image/issues/2238

### DIFF
--- a/src/compose.rs
+++ b/src/compose.rs
@@ -239,8 +239,8 @@ where
 /// use imageproc::definitions::Image;
 /// use image::LumaA;
 ///
-/// let bottom = Image::from_pixel(4, 4, LumaA([0, 255]));
-/// let top = Image::from_pixel(4, 4, LumaA([0, 255]));
+/// let bottom = Image::from_pixel(4, 4, LumaA([0u8, 255]));
+/// let top = Image::from_pixel(4, 4, LumaA([255u8, 0]));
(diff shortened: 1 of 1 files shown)


src/compose.rs
```diff
@@ -160,3 +162,117 @@ where
...
+pub fn replace<P>(bottom: &Image<P>, top: &Image<P>, x: u32, y: u32) -> Image<P>
+where
+    P: Pixel,
+{
+    let mut bottom = bottom.clone();
+    replace_mut(&mut bottom, top, x, y);
+    bottom
+}
+#[doc=generate_mut_doc_comment!("replace")]
+pub fn replace_mut<P>(bottom: &mut Image<P>, top: &Image<P>, x: u32, y: u32)
```
theotherphil, line 211: Isn’t this just https://docs.rs/image/latest/image/trait.GenericImage.html#method.copy_from with a different name?

theotherphil: I guess the error type on the `GenericImage` function is a bit clunky.

ripytide: I didn't see `copy_from()`, I think the functions are similar but being based on `GenericImage` may also have performance issues as we've found previously.

Also technically it is an image processing operation so belongs here more than in `image` but I don't feel too strongly about that.

PR #669, opened 2024-06-12
Add rotation module and functions

ripytide:
As a part of https://github.com/image-rs/image/issues/2238


theotherphil:
Thanks! These should go alongside the existing rotation functions in the (clunkily named) geometric_transformation module.

ripytide:
I've moved them across :+1:

ripytide:
Do you think we should move the `flip` functions to `geometric_transformations` as well since they are like reflections?

theotherphil:
Let’s leave them for now and bike shed the module structure with whatever functions we’ve got before the next release :-)

src/geometric_transformations.rs
```diff
@@ -317,6 +317,146 @@ where
...
+///
+/// let rotated = rotate90(&image);
+///
+/// assert_eq!(rotated, gray_image!(
+///     0, 0, 3, 1;
+///     0, 0, 4, 2;
+///     0, 0, 0, 0;
+///     0, 0, 0, 0));
+/// ```
+pub fn rotate90<P>(image: &Image<P>) -> Image<P>
```
theotherphil, line 341: I don’t feel strongly either way but it’s a bit unfortunate to use angles in degrees in the names of these functions when angles everywhere else in this file (/crate) are in radians. Can’t think of better names off the top of my head though.

ripytide: As some prior art, opencv just has a single `rotate()` function which takes an enum that looks like `ROTATE_90_CLOCKWISE`

ripytide: Maybe we should do the same and use a single function with an enum, but I don't feel particularly strongly either way.

theotherphil: I wondered the same thing. I’ll just merge as is for now. Thanks.

PR #670, opened 2024-06-22
Expand `stetch_contrast()` to color images also using `u8`

ripytide:
Ideally we'd want to use the `Enlargeable` trait currently private inside `image` to make this work for all types rather than just `u8`, but this is still an improvement that should be good enough for https://github.com/image-rs/image/issues/2238.

### DIFF
--- a/src/contrast.rs
+++ b/src/contrast.rs
@@ -2,13 +2,13 @@
 
 use std::cmp::{max, min};
 
-use image::{GrayImage, Luma};
+use image::{GrayImage, Luma, Pixel};
 #[cfg(feature = "rayon")]
 use rayon::prelude::*;
 
-use crate::definitions::{HasBlack, HasWhite};
+use crate::definitions::{HasBlack, HasWhite, Image};
 use crate::integral_image::{integral_image, sum_image_pixels};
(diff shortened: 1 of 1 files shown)


ripytide:
It seems we need to bump the minimal rust version now since one of our dependencies, I think `bitstream-io` now requires `v1.79`

theotherphil:
Thanks! What impact does this have on function run time for the greyscale case?

ripytide:
Looks like impact:
Before PR:
```
test contrast::benches::bench_stretch_contrast                                        ... bench:         660.08 ns/iter (+/- 2.24)
test contrast::benches::bench_stretch_contrast_mut                                    ... bench:           0.25 ns/iter (+/- 0.00)
```
After PR:
```
test contrast::benches::bench_stretch_contrast                                        ... bench:         658.86 ns/iter (+/- 7.36)
test contrast::benches::bench_stretch_contrast_mut                                    ... bench:           0.25 ns/iter (+/- 0.00)
```

theotherphil:
Thanks!

(Aside: it looks like the _mut variant of that benchmark is optimising out the function and so not actually running .)

PR #680, opened 2024-08-04
Fix fast corner 9 algorithm

LMJW:
Simple bug fix for FAST corner 9 algorithm.

Currently the fast corner 9 has a bug which causing some of the corner cannot be found. It looks the coordinate (P4& P8) was not mapped correctly in fast algorithm. 

Below are the corner find without bug fix, some of the corner cannot be identified.  
corner1

With the fix, all the corners can be found.
corner2


LMJW:
@ripytide Pls take a look when you have time and let me know if you have any comments :)

ripytide (approved):
It looks as though p4 and p8 were the wrong way round. Considering this code looks nearly 9 years old that's a great catch!

theotherphil:
Thanks! It would be good to have some tests for this too.

PR #682, opened 2024-08-12
handle polygon draw case where start point equals end point

fs-99:
Thank you for the nice library, I'm using it mostly for basic draw calls.
I stumbled upon this when drawing shapes with randomly generated points, where I don't really care if this backend adds the endpoint for the last line or if I randomly generated the same end point as the start point.

### DIFF
--- a/src/drawing/polygon.rs
+++ b/src/drawing/polygon.rs
@@ -102,17 +102,7 @@ where
         return;
     }
     if poly.len() < 2 {
-        panic!(
-            "Polygon only has {} points, but at least two are needed.",
-            poly.len(),
-        );
-    }
-    if poly[0] == poly[poly.len() - 1] {
-        panic!(
-            "First point {:?} == last point {:?}",
-            poly[0],
-            poly[poly.len() - 1]
-        );
+        panic!("Polygon only has 1 point, but at least two are needed.");
     }
     for window in poly.windows(2) {
         crate::drawing::draw_line_segment_mut(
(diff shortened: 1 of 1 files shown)


fs-99:
I forgot to change the docs: should I just remove the point about open paths or should I say "This supports both open and closed paths, i.e. the first and last points can be the same but don't have to."

src/drawing/polygon.rs
```diff
@@ -123,8 +113,10 @@ where
         );
     }
     let first = poly[0];
```
ripytide, line 115:
```suggestion
    let first = poly.first().unwrap();
```

src/drawing/polygon.rs
```diff
@@ -123,8 +113,10 @@ where
...
-    let last = poly.iter().last().unwrap();
-    crate::drawing::draw_line_segment_mut(canvas, (first.x, first.y), (last.x, last.y), color);
+    let last = poly[poly.len() - 1];
```
ripytide, line 116:
```suggestion
    let last = poly.last().unwrap();
```

src/drawing/polygon.rs
```diff
@@ -174,7 +159,9 @@ where
     y_max = max(0, min(y_max, height as i32 - 1));
 
     let mut closed: Vec<Point<i32>> = poly.to_vec();
-    closed.push(poly[0]);
+    if poly[0] != poly[poly.len() - 1] {
+        closed.push(poly[0]);
+    }
```
ripytide, line 164:
```suggestion
    let first = poly.first().unwrap();
    let last = poly.last().unwrap();

    if first != last {
        closed.push(first);
    }
```

ripytide (approved):
Looks useful, I would change the docs to say that it supports open and closed paths and that if the last point is not equal to the first then the path is closed anyway.

src/drawing/polygon.rs
theotherphil: Switching from indexing to first/last is more verbose and less idiomatic.

theotherphil:
Seems like reasonable behaviour if clearly documented. Thanks, I’ll merge after the docs are updated to explain the new behaviour.

fs-99:
Done. (and @ripytide I gave it the names, the length is already checked if that was your point)

theotherphil:
Thanks!

PR #684, opened 2024-09-05
New feature: Flood-fill

tkallady:
Similar to bucket tool in MS-PAINT.
I've just taken the algorithm from here:
https://en.wikipedia.org/wiki/Flood_fill#Span_filling

I'm fairly new to rust and didn't know how to do generic Pixel equality without first converting to rgba pixel, perhaps someone knows how to do this better?:
https://github.com/image-rs/imageproc/blob/d2a187ceecdde5d7e4a4526531d7c0c909dcd227/src/drawing/fill.rs#L61


src/drawing/fill.rs
```diff
@@ -0,0 +1,62 @@
+use image::{GenericImage, Pixel};
+
+/// Equivalent to bucket tool in MS-PAINT
+/// Performs 4-way flood-fill based on this algorithm: https://en.wikipedia.org/wiki/Flood_fill#Span_filling
+pub fn flood_fill<I>(image: &mut I, x: u32, y: u32, fill_with: I::Pixel)
+where
+    I: GenericImage,
```
ripytide, line 7: We tend to prefer using the `Image` type rather than `GenericImage` for performance reasons. Also methods that take a `&mut` image should normally end with `_mut`, see other similar methods for examples on creating both variants and using the doc macro. I'm imagining something like this
```suggestion
pub fn flood_fill_mut<P>(image: &mut Image<P>, x: u32, y: u32, fill_with: P)
where
    P: Pixel,
```

tests/regression.rs
```diff
@@ -748,6 +749,25 @@ fn test_draw_filled_ellipse() {
...
+    draw_hollow_ellipse_mut(&mut image, (100, 100), 50, 50, red);
+    draw_hollow_ellipse_mut(&mut image, (50, 100), 40, 90, blue);
+    draw_hollow_ellipse_mut(&mut image, (100, 150), 80, 30, green);
+    draw_hollow_ellipse_mut(&mut image, (150, 150), 100, 60, blue);
+
+    flood_fill(&mut image, 120, 120, red);
+
+    compare_to_truth_image(&image, "flood_filled_shape.png");
+}
+
```
ripytide, line 770: Awesome test!

src/drawing/fill.rs
```diff
@@ -0,0 +1,62 @@
...
+                stack.push((x, x1 - 1, y + dy, dy))
+            }
+            if x1 - 1 > x2 {
+                stack.push((x2 + 1, x1 - 1, y - dy, -dy))
+            }
+            x1 = x1 + 1;
+            while x1 < x2 && !inside(image, x1, y, target) {
+                x1 = x1 + 1
+            }
+            x = x1
```
ripytide, line 44: Is there a way to write this with less mutable state as that makes it very hard to read. Maybe using iterator over pixels with a filter maybe?

src/drawing/fill.rs
```diff
@@ -0,0 +1,62 @@
...
+    I: GenericImage,
+{
+    if x < 0 || y < 0 {
+        return false;
+    }
+    let x = x as u32;
+    let y = y as u32;
+    let (width, height) = image.dimensions();
+    //TODO: Compare pixel equality without conversion to rgba
+    x < width && y < height && image.get_pixel(x, y).to_rgba() == target_color.to_rgba()
```
ripytide, line 61: Try adding a `PartialEq` bound onto the pixel type `P` and then you will be able to do `image.get_pixel(x, y) == target_color`. I would probable also rename `target_color` to target_pixel since not all pixel types are colors.

tkallady: Thanks, updated accordingly

src/drawing/fill.rs
tkallady: Done

src/drawing/fill.rs
tkallady: I agree it is ugly but I'm not sure that's avoidable with this particular algorithm.
It is copied verbatim from pseudocode on the wikipedia page so at least there is some explanation there.
There are simpler algorithms that would be more readable but less performant. The most elegant one (recursive) causes a stack overflow.

src/drawing/fill.rs
```diff
@@ -1,12 +1,23 @@
...
-    I: GenericImage,
+    P: Pixel + PartialEq,
 {
-    let target = image.get_pixel(x, y);
+    let mut filled_image = image.clone();
+    flood_fill_mut(&mut filled_image, x, y, fill_with);
+    filled_image
+}
+
+#[doc=generate_mut_doc_comment!("draw_line_segment")]
```
ripytide, line 15: You need to update this to the same name as the non-mut method

tests/regression.rs
```diff
@@ -763,8 +762,10 @@ fn test_draw_flood_filled_shape() {
     draw_hollow_ellipse_mut(&mut image, (100, 150), 80, 30, green);
     draw_hollow_ellipse_mut(&mut image, (150, 150), 100, 60, blue);
 
-    flood_fill(&mut image, 120, 120, red);
+    let filled_image = flood_fill(&image, 120, 120, red);
+    compare_to_truth_image(&filled_image, "flood_filled_shape.png");
 
+    flood_fill_mut(&mut image, 120, 120, red);
```
ripytide, line 768: Not really any need to test both variants since they are the same method underneath.

src/drawing/fill.rs
ripytide: Ah ok, if it's copied from the wikipedia then I'm satisfied with the correctness.

ripytide:
Apart from the two new nits, I'm happy for this PR to be merged.

tkallady:
All done. Thanks.

cospectrum:
for `clippy lint` you can merge `main`, for `doc lint` just fix url

tkallady:
CI should pass now.

cospectrum:
Actually I think `clippy` will break with each new lint until they run out of ideas.

theotherphil:
Thanks!

PR #686, opened 2024-09-13
Update `CI`

cospectrum:
This PR fixes a new clippy [lint](https://rust-lang.github.io/rust-clippy/master/index.html#/too_long_first_doc_paragraph) and adds a [job](https://github.com/rhysd/actionlint) to check the gh-workflows syntax

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,15 @@
+name: Lint GitHub Actions workflows
+on: [push, pull_request]
+
+jobs:
+  actionlint:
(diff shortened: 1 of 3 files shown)


ripytide:
Rather than globally disable the lint why not simply fix the too long first doc paragraph? I think the first sentence would be fine as is we could just add a `<cr>` after `Hue rotate the supplied image in place.`

The actionlint looks cool I never new that project existed!

cospectrum:
Allowed because there are false positives

theotherphil:
Thanks!

Issue #687, opened 2024-09-21
Text rendered at wrong Y position


SpriteOvO:
Minimal reproducible

```rust
use ab_glyph::FontRef;
use imageproc::{
    drawing::*,
    image::{Rgb, RgbImage},
    rect::Rect,
};

fn main() {
    let mut image = RgbImage::new(200, 200);

    let font = FontRef::try_from_slice(include_bytes!("../DejaVuSans.ttf")).unwrap();
    let white = Rgb([255, 255, 255]);

    let scale = 64.;
    let text = "hello";
    let (w, h) = text_size(scale, &font, text);

    draw_text_mut(&mut image, white, 0, 0, scale, &font, text);
    draw_hollow_rect_mut(&mut image, Rect::at(0, 0).of_size(w, h), white);

    image.save("./output.png").unwrap();
}
```

Rendered

![image](https://github.com/user-attachments/assets/0caf10ce-869f-4bda-82b6-a2ef1ccdd5a5)

This issue would be more obvious if we used other fonts or languages.

For example, `NotoSansCJK-Regular.ttc`

![image](https://github.com/user-attachments/assets/ddfddacd-8bc3-48ad-8ad8-383135e6517a)
![image](https://github.com/user-attachments/assets/d1cf1a37-86d4-456b-b1de-78c379c03820)

cospectrum:
I think It's not the wrong Y position, it's the wrong `text_size`

cospectrum:
It looks like glyph "bearing" at the top. But `text_size` is broken for sure!

PR #689, opened 2024-10-19
fix `text_size`

cospectrum:
First I tried `h = h.max(bb.min.y + bb.height());`. But then I think I found a better solution.
Looks good when using English and Japanese text. And some useful distance at the bottom of the text. You can try on my branch.
Relates: https://github.com/image-rs/imageproc/issues/687


cospectrum:
I think `clippy` should be optional because we use the `nightly` version to get the latest features and improve the code. 
But there are no "allowed to fail" jobs on `github-actions`, thus, we will live with the red status

cospectrum:
It's not that bad. I see that it discovered unnecessary lifetimes and stuff. We just have to fix it every couple of months.

cospectrum:
Maybe we can add more versions to clippy matrix
```yaml
matrix:
    toolchain: [nightly]
```

cospectrum:
> Maybe we can add more versions to clippy matrix
> 
> ```yaml
> matrix:
>     toolchain: [nightly]
> ```

No, stable `clippy` doesn't understand test code with nightly features

cospectrum:
I left text rendering unchanged, didn't find any fatal errors.

`rect.height() == font.height()`.

font

theotherphil:
Thanks!

it would be nice to extend the text regression test to include the issue this is fixing: https://github.com/image-rs/imageproc/blob/54573dabc28e0fdd8f978843e253d06bc763cf14/tests/regression.rs#L810

cospectrum:
Ofc. I will write proptests to ensure that text is always rendered inside this rectangle.

cospectrum:
And now I have found a counterexample for the X position in `draw_text` itself (first `j`).
```rs
let x = 62;
let y = 21;
let scale = 78.0932;
let text = "jnenuoltqjylzcbvcc";
```
font

cospectrum:
I wonder if this is a bug in the `ab_glyph` bbox or a rounding error in the `imageproc`?

cospectrum:
Maybe the origin of the first glyph? its (0, font.ascent())

cospectrum:
I tried to fix it, but it requires replacing the `text.png` sample from the regression tests.

cospectrum:
`proptest` passed for english text with numbers

theotherphil:
The test image can be regenerated by running

`REGENERATE=1 cargo test`

cospectrum:
Are you familiar with OpenType glyphs? 
I didn't plan to change `draw_text` at all. But now I'm shifting all glyphs to the right by `abs(h_side_bearing)` of the first glyph, because some glyphs go beyond of `Rect.at(x, y)`, and it's not just a floating point error.

cospectrum:
I think that despite moving to the right, all relative distances will remain the same as in the previous implementation.

cospectrum:
For example, in the previous implementation, the drawing of `j` with `(x, y) = (0, 0)` will go beyond the image.

cospectrum:
We can go the other way. We can keep the starting point at 0, that is, no change in the `draw_text` implementation. And we can document that in `draw_text`. 
Instead, we will change my new tests so that the "checked" rectangle itself takes into account the negative bearing.
What do you think?

cospectrum:
We can also try to add the horizontal `bearing` to the origin only if it is negative.

theotherphil:
I’m not familiar with OpenType.

Any of these options seems fine - let’s go with relaxing the test to allow for negative side bearings if you have no preference.

cospectrum:
Sure, one moment

cospectrum:
Done

cospectrum:
Finally, almost 1 day is spent.

cospectrum:
I added `+1` because otherwise `rect.contains` will fail

theotherphil:
Thanks a lot!

PR #693, opened 2024-11-18
display_multiple_images with different titles

Icekey:
I changed display_multiple_images so that the windows can have different titles. In my use case i have similar looking images that i want to compare. Not having the ability to define different window titles makes this task very confusing.

The implementation repeats the last title if more images than titles are passed to the function. If no title is passed then all windows have no title.


cospectrum:
CI will pass after rebase I guess. But I think the number of images should be equal to the number of titles so that we don't do some weird magic. This can be checked with `assert_eq`. 
Others can share their opinion.

Issue #694, opened 2024-11-21
Detailen error for generic type parameter of find_contours_with_threshold


sakird:
I am learning Rust and tried to use find_contours_with_threshold function that I have asked ChatGPT to prepare. 

Calling:
`
    let contours = find_contours_with_threshold::<u8>(&binary_edges, 0u8);
`

Image dimensions were too big for u8. And since I am new to Rust I have been trying to find the cause of this error. I have finally managed to find it, here:

`
contour_points.push(Point::new(cast(pos3.x).unwrap(), cast(pos3.y).unwrap()));
`

After changing that dimension to u32, it worked. 
This cast panics if image width or height is bigger than max of u8. Would it be OK to update this code like this:

`
contour_points.push(Point::new(
    cast(pos3.x).expect("Generic data type should be bigger than width."),
    cast(pos3.y).expect("Generic data type should be bigger than height."),
));
`

If this is acceptable, I can send a PR.

ripytide:
That seems like an improvement to me. Since this is a user facing error then having an error message is better than nothing. Thought I might make the error message : `Generic Type Parameter T should be bigger than width`/`Generic Type Parameter T should be bigger than height`.

sakird:
Then I will send a PR. Thanks.

PR #695, opened 2024-11-22
Detailed error message for generic type parameter of find_contours_with_threshold (#694)

sakird:
Below code does not help developers in case of an error with the size set using generic type parameter:

`
contour_points.push(Point::new(cast(pos3.x).unwrap(), cast(pos3.y).unwrap()));
`

To help developers understand the error easily, code is updated with the error message like below:

`
contour_points.push(Point::new(
    cast(pos3.x).expect("Generic Type Parameter T should be bigger than width."),
    cast(pos3.y).expect("Generic Type Parameter T should be bigger than height."),
));
`

### DIFF
--- a/src/contours.rs
+++ b/src/contours.rs
@@ -147,8 +147,10 @@ where
                     let mut pos3 = curr;
 
                     loop {
-                        contour_points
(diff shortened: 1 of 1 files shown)


cospectrum:
Relates: https://github.com/image-rs/imageproc/issues/694

Issue #696, opened 2024-12-04
Kapur's algorithm


oqpvc:
Hi there,
I've implemented Kapur's algorithm for contrast threshold detection (https://github.com/oqpvc/autograder/commit/67d1f92417eb3c5b836e03e77cb31e5df602874d). This might be a reasonable addition to imageproc - do you want a pull request?

theotherphil:
That would be a nice addition to the library - thanks!

PR #699, opened 2024-12-06
Implement Kapur's algorithm for binary thresholding (#696)

oqpvc:
Kapur's algorithm uses an entropy approach to find a good segmentation threshold for gray images.

This implementation of Kapur's algorithm sticks closely to the notation of the original article (https://doi.org/10.1016/0734-189X(85)90125-2), so it is much easier to check for correctness. We also avoid a double loop, so this only needs 4*256 computations instead of 256^2.

### DIFF
--- a/src/contrast.rs
+++ b/src/contrast.rs
@@ -114,54 +114,56 @@ pub fn kapur_level(img: &GrayImage) -> u8 {
     // straight from the article.
     let hist = histogram(img);
     let histogram = &hist.channels[0];
+    const N: usize = 256;
 
     let total_pixels = (img.width() * img.height()) as f64;
 
     // The p_i in the article. They describe the probability of encountering
     // gray-level i.
-    let mut p = [0.0f64; 256];
-    for i in 0..=255 {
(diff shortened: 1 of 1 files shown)


cospectrum:
Why not replace `f64::EPSILON` with `0.0`?

theotherphil:
Thanks! Please can you make the change suggested by clippy (not obviously an improvement, but not worse and this is easier than ad-hoc disabling of lints) and then I’ll merge.

cospectrum:
I will make a PR to lock the nightly channel for `clippy`

oqpvc:
> Thanks! Please can you make the change suggested by clippy (not obviously an improvement, but not worse and this is easier than ad-hoc disabling of lints) and then I’ll merge.

I can absolutely do that, but just for the record: The clippy change is about code that has nothing to do with the PR. You still want me to push it here?

theotherphil:
Ah, I hadn’t spotted that. Fixed by @cospectrum in https://github.com/image-rs/imageproc/pull/700.

Thanks for the new feature.

Issue #703, opened 2025-01-03
Feature request: image hashing


cospectrum:
I'm interested in seeing perceptual hashing in the upstream. Do I have a green light?

theotherphil:
Sounds like a nice feature to have - thanks!

PR #709, opened 2025-01-08
add `pHash`

cospectrum:
Relates: https://github.com/image-rs/imageproc/issues/703


cospectrum:
I know there are more optimizations. It's a starting point.

Cargo.toml
```diff
@@ -54,6 +55,9 @@ features = ["property-testing", "katexit"]
 opt-level = 3
 debug = true
 
+[profile.dev]
+opt-level = 2
```
theotherphil, line 59: What’s the reason for this change? It would be a shame to make debug builds take even longer.

src/imghash/mod.rs
```diff
@@ -0,0 +1,11 @@
+//! [Perceptual hashing] algorithms for images.
+//!
+//! [Perceptual hashing]: https://en.wikipedia.org/wiki/Perceptual_hashing
+
```
theotherphil, line 4: The rest of this library uses “image” over “img” for module and function names - please can you rename this to image_hash or maybe perceptual_hashing.

theotherphil (commented):
Overall this looks great, thanks a lot. I’ll merge after the two comments have been resolved.

Cargo.toml
cospectrum: To speed up proptests

cospectrum: Changed to `opt-level = 1`

src/imghash/mod.rs
cospectrum: renamed to `image_hash`

cospectrum:
I updated MSRV because `image-webp@0.2.1` requires rustc `1.80.1`.

theotherphil:
Nice, thanks!

PR #715, opened 2025-03-26
Add a new mask canvas

wani3327:
This adds the  struct to the crate, providing a simple way to easily draw on a specific part of an image.


cospectrum:
For future reference, CI issues related to `sdl2` cmake build, as far as I remember

PR #763, opened 2026-03-18
add regression test for draw_text_mut (#244)

### LINKED ISSUE #244: Add regression tests for text drawing functions

JakobZahn:
* add test_draw_text_outside_bounds() to cover the case of #174

Keyword: fix #244

### DIFF
--- a/tests/regression.rs
+++ b/tests/regression.rs
@@ -1003,6 +1003,27 @@ fn test_draw_text() {
 
 #[test]
 #[cfg(feature = "text")]
+fn test_draw_text_outside_bounds() {
+    use proptest::char::range;
+
+    let font_bytes = include_bytes!("data/fonts/DejaVuSans.ttf");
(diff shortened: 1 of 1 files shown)


JakobZahn:
Hello, I am new to contributing and rust.

JakobZahn:
The issue itself seems to be mostly covered already as there are two regression tests for draw_text_mut in regression.rs. So this might already fix #244.
However, the special regression test for issue https://github.com/image-rs/imageproc/pull/174 was missing and I tried to add it. Unfortunately, I could not reproduce the original issue and the respective code snippet in #174 is no longer available, so that might make my test worthless. I tried checking out the old version but it is so long ago that I ran into several problems which seemed not worth the effort.

My test does not compare to a truth image because the initial issue in #174 was a panic.

PR #766, opened 2026-03-25
Bilateral speedup

CattleProdigy:
At \<dayjob\>  we noticed a pretty serious regression in performance in the bilateral filter when upgrading from 0.25 to 0.26. So I worked on speeding it up.

Not allocating more than necessary is by far the most significant improvement. The `cartesian_product` helper has some noticeable overhead (10s of ms). I also tried a well-known approximation for the `exp()`, I'm a little more skeptical of adding that one though. LMK

```
Baseline:
gray 500x500 r=5:
  median:   901.39ms  mean:   901.54ms  min:   900.65ms  max:   903.81ms
rgb  500x500 r=5:
  median:   939.59ms  mean:   940.08ms  min:   938.23ms  max:   942.73ms

Skip allocations and cartesian_product:
gray 500x500 r=5:
  median:   101.20ms  mean:   101.27ms  min:   101.12ms  max:   101.59ms
rgb  500x500 r=5:
  median:   155.27ms  mean:   155.30ms  min:   155.15ms  max:   155.53ms

exp approximation (+ previous optimizations)
gray 500x500 r=5:
  median:    74.99ms  mean:    75.01ms  min:    74.90ms  max:    75.14ms
rgb  500x500 r=5:
  median:   123.16ms  mean:   123.19ms  min:   122.96ms  max:   123.66ms
```

### DIFF
--- a/src/filter/bilateral.rs
+++ b/src/filter/bilateral.rs
@@ -20,10 +20,19 @@ pub struct GaussianEuclideanColorDistance {
     sigma_squared: f32,
 }
 impl GaussianEuclideanColorDistance {
-    /// Creates a new [`GaussianEuclideanColorDistance`] using a given sigma value.
+    /// Creates a new [`GaussianEuclideanColorDistance`] using a given sigma value, which
+    /// must be positive.
     ///
     /// Internally, this is stored as sigma squared for performance.
+    ///
+    /// # Panics
+    ///
+    /// 1. If `sigma <= 0`
     pub fn new(sigma: f32) -> Self {
+        assert!(
+            sigma > 0.0,
+            "GaussianEuclideanColorDistance sigma must be positive"
+        );
         GaussianEuclideanColorDistance {
             sigma_squared: sigma.powi(2),
         }
@@ -55,7 +64,7 @@ where
 /// * `radius` - The radius of the kernel used for the filtering. 0 -> 1x1, 1 -> 3x3, 2 -> 5x5, 3
 ///     -> 7x7, etc..
 /// * `spatial_sigma` - Standard deviation for euclidean spatial distance. A larger value results in
-///     averaging of pixels with larger spatial distances.
+///     averaging of pixels with larger spatial distances. Must be positive.
 /// * `color_distance` - A type which implements [`ColorDistance`]. This defines the metric used to
 ///     define how different two pixels are based on their colors. Common examples may include simple
 ///     absolute difference for greyscale pixels or cartesian distance in the CIE-Lab color space
@@ -77,6 +86,7 @@ where
 /// 2. If `image.height() > i32::MAX as u32`.
 /// 3. If `image.width() == 0`
 /// 4. If `image.height() == 0`
+/// 5. If `spatial_sigma <= 0`
 ///
 /// # Examples
 ///
@@ -112,6 +122,7 @@ where
     assert!(!image.height() > i32::MAX as u32);
     assert_ne!(image.width(), 0);
     assert_ne!(image.height(), 0);
+    assert!(spatial_sigma > 0.0, "spatial_sigma must be positive");
 
     let radius = i16::from(radius);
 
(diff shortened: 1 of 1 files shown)


src/filter/bilateral.rs
```diff
@@ -177,6 +177,29 @@ fn gaussian_weight(x_squared: f32, sigma_squared: f32) -> f32 {
...
+/// Based on: N. Schraudolph, "A Fast, Compact Approximation of the Exponential Function",
+/// Neural Computation 11(4), 1999.
+///
+/// Exploits the IEEE 754 float layout: reinterprets `a * x + b` as the bit pattern of an f32,
+/// where `a` and `b` are chosen so the exponent and mantissa fields approximate `exp(x)`.
+///
+/// Valid for `x` in roughly `[-87, 0]`. Returns 0 for `x < -87` (where true exp underflows).
+/// Maximum relative error is ~4% in the range used by the bilateral filter.
+#[inline]
+fn fast_exp_negative(x: f32) -> f32 {
```
cospectrum, line 207: We should add `assert` (or at least `debug_assert`) to check the `x <= 0` condition.

cospectrum:
could you please add unit-tests (or better proptests) to check that the error of fast_exp is less than 4%?

src/filter/bilateral.rs
cospectrum: you can benchmark the version with `assert`

CattleProdigy:
> could you please add unit-tests (or better proptests) to check that the error of fast_exp is less than 4%?

See changes here: https://github.com/image-rs/imageproc/pull/766/changes/c0bbe0133ab957de2c3aab4a24a8a83d5d898ecf

I tested all negative f32 values until -87 where `f32::exp` underflows anyway.

src/filter/bilateral.rs
```diff
@@ -111,30 +110,31 @@ where
...
         debug_assert!(image.in_bounds(x, y));
         // Safety: `Image::from_fn` yields `col` in [0, width) and `row` in [0, height).
         let center_pixel = unsafe { image.unsafe_get_pixel(x, y) };
 
-        let window_len = 2 * radius + 1;
-        let weights_and_values = window_range
-            .clone()
-            .cartesian_product(window_range.clone())
-            .map(|(w_y, w_x)| {
+        let mut channel_sums = [0f32; 4];
```
cospectrum: Add `assert!(P::CHANNELS_COUNT as usize <= 4);` at the beginning of the function.
And probably better to use `const MAX_CHANNELS: usize = 4;`

cospectrum (changes requested):

src/filter/bilateral.rs
CattleProdigy: e2ba0405aac31909ec6505354cfb8ef75739556b

The cost of using `assert!` is 5-6%

src/filter/bilateral.rs
CattleProdigy: c4d2044746c96efb46bc73fa75aaf094578cdf4f

One day we'll get `#![feature(generic_const_exprs)]` or whatever lets us do this statically.

src/filter/bilateral.rs
CattleProdigy: hmm unit tests are proptesting with sigmas that are negative, zero, or close enough to zero that they cause NaNs. This is an existing issue, the new assert just finds it. 

I can add some more `asserts!` and fix the sigma values in the proptests. Very small values of sigma will be tricky since they're positive and thus meet the theoretical criterion but will produce NaNs when used in division. I can allow `NaN` in the `debug_assert!`. 

`NaN` is currently allowed to filter in currently. But my changes introduce a difference in behavior.

`f32::NAN.exp() = NAN`
vs.
`fast_exp_negative(NAN) = 0` 

For filtering operations that eventually lead back to an integer type, you'll get a 0 anyway due to casting rules.  I've checked that with `debug_assert!(AsPrimitive::<i32>::as_(f32::NAN) == 0);`. But with a floating pixel type we'll get 0 instead of `NaN`. 

It's not clear to me if this is worth caring about. Only a narrow range of unrealistic values could trigger this. 

Two other alternatives:

- Disallow NaNs in the `debug_assert!`, add some epsilon to the proptest sigma to keep it away from zero. 
- `assert!` that sigma must be some epslion greater than 0 to avoid NaNs. It's hard to know what that epsilon will be since it'll depend on e.g. the scale of pixel values.

CattleProdigy: addressed here: 416f71a13ce7229eb2e7cfde55ad84c3ca6cedcf

PR #777, opened 2026-04-11
Fix u32 overflow in Kernel::new dimension check

### LINKED ISSUE #776: u32 overflow in `Kernel::new` leads to OOB read in `filter()`
AI disclosure: this issue was found via Claude Code running Claude Opus 4.6.

## Summary

`Kernel::new` validates its dimensions with `assert!(width * height == data.len() as u32)`. In release mode, the `width * height` multiplication can silently overflow u32, causing the assertion to pass with a data slice far smaller than the kernel's logical dimensions. When this malformed kernel is passed to `filter()`, the inner loop calls `kernel.get_unchecked()` with an out-of-bounds index, producing undefined behavior.

The entire chain uses only safe public API. No `unsafe` is required in user code.

## Minimal reproducer

```shell
cargo init --name repro /tmp/kernel-overflow-repro
cd /tmp/kernel-overflow-repro

cargo add --git https://github.com/image-rs/imageproc.git --rev 1e5197fde065d66eb1ef44863f0e6b77d3ffc754 --no-default-features
cargo add image@=0.25.10 --no-default-features
# alternatively, use the latest crates.io release: cargo add imageproc@0.26.1 --no-default-features

cat > src/main.rs <<'EOF'
use image::Luma;
use imageproc::filter::filter;
use imageproc::kernel::Kernel;

fn main() {
    // 2 * 2_147_483_649 = 4_294_967_298, which overflows u32 to 2.
    // Kernel::new's assert `width * height == data.len() as u32` sees 2 == 2 and passes.
    let data = [0.0f32; 2];
    let kernel = Kernel::new(&data, 2, 2_147_483_649);

    let image = image::ImageBuffer::from_pixel(1, 1, Luma([128u8]));

    // filter() iterates k_y in 0..2_147_483_649, k_x in 0..2.
    // On the 3rd iteration (k_y=1, k_x=0), kernel.get_unchecked(0, 1)
    // computes index = 1*2 + 0 = 2 — OOB on the 2-element data slice.
    let _: image::ImageBuffer<Luma<f32>, Vec<f32>> = filter(&image, kernel, |x| x);
}
EOF

RUSTFLAGS="-Coverflow-checks=no -Cdebug-assertions=no" cargo +nightly miri run
```

The `-Coverflow-checks=no -Cdebug-assertions=no` flags simulate the release profile
(`--release` enables both). The result:

```
error: Undefined Behavior: `assume` called with `false`
  --> /home/paolobarbolini/.cargo/git/checkouts/imageproc-15e7df7e968bb8fb/1e5197f/src/kernel.rs:60:14
   |
60 |             *self.data.get_unchecked(at)
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
   |
   = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
   = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
   = note: stack backtrace:
           0: imageproc::kernel::Kernel::<'_, f32>::get_unchecked
               at /home/paolobarbolini/.cargo/git/checkouts/imageproc-15e7df7e968bb8fb/1e5197f/src/kernel.rs:60:14: 60:41
           1: imageproc::filter::filter::<image::Luma<u8>, f32, {closure@src/main.rs:16:77: 16:80}, image::Luma<f32>>
               at /home/paolobarbolini/.cargo/git/checkouts/imageproc-15e7df7e968bb8fb/1e5197f/src/filter/mod.rs:57:25: 57:69
           2: main
               at src/main.rs:16:54: 16:83
```

paolobarbolini:
Widen width and height to u64 before multiplying so that overflowing dimensions are correctly rejected instead of silently passing the assertion and later causing OOB reads in filter().

Fixes #776

### DIFF
--- a/src/kernel.rs
+++ b/src/kernel.rs
@@ -23,7 +23,8 @@ where
     /// 2. If `width * height != data.len() as u32`.
     pub const fn new(data: &'a [K], width: u32, height: u32) -> Kernel<'a, K> {
         assert!(width > 0 && height > 0, "width and height must be non-zero");
-        assert!(width * height == data.len() as u32);
+        // Widen to u64 to prevent u32 overflow from bypassing the assertion.
+        assert!(width as u64 * height as u64 == data.len() as u64);
         Kernel {
             data,
             width,
@@ -90,3 +91,17 @@ pub const ROBERTS_VERTICAL_2X2: Kernel<'static, i32> = Kernel::new(&[0, 1, -1, -
 
 /// The laplacian 3x3 kernel.
 pub const LAPLACIAN_3X3: Kernel<'static, i16> = Kernel::new(&[0, 1, 0, 1, -4, 1, 0, 1, 0], 3, 3);
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    #[should_panic]
+    fn kernel_new_rejects_overflowing_dimensions() {
+        // 2 * 2_147_483_649 overflows u32 to 2, matching data.len().
+        // The widened u64 check must catch this.
+        let data = [0.0f32; 2];
+        let _ = Kernel::new(&data, 2, 2_147_483_649);
+    }
+}


Shnatsel:
Would you also publish a new release with this change to crates.io? It would be nice to get the soundness fix shipped to users.

PR #778, opened 2026-04-28
Fix out-of-bounds reads in BRIEF

Shnatsel:
The issue was discovered by Claude Opus 4.7 and reported privately earlier.


Shnatsel:
The proptest failure is unrelated to the changes in this PR. It must be a pre-existing issue with a flaky proptest.

PR #793, opened 2026-06-26
Guard oriented_fast edge_radius against u32 underflow (DoS)

mahirhir:
`oriented_fast` does an unchecked `edge_radius` subtraction that underflows `u32`, panicking in debug and looping ~1.8e19 times in release (DoS) for an unvalidated radius. #777/#778 added `saturating_sub` hardening to the sibling APIs; `oriented_fast` was left unmirrored. The fix applies the same `saturating_sub`. 553 tests pass, 0 regression.


mahirhir:
Closing this. I can't compile-check it locally and CI is red on it, so I don't want to leave an unverified change open. Sorry for the noise.

PR #798, opened 2026-07-13
Bilateral speedup

CattleProdigy:
Even after the last round of performance improvements to the bilateral filter there was still a pretty big performance drop starting after 0.25. It turned out to mostly because of confusing the window-radius vs window-size argument change. Nonetheless, I kept making performance improvements.

This PR has two more improvements:

1. Introduction of a LUT-based color-distance implementation for `u8`-subpixel pixel types. This LUT implementation has twice the memory footprint, but is faster since it avoids an absolute value.
2. Separate code paths for pixels in the image boundary which requires clamping and pixel in the middle which don't. Clamping code showed up as a fairly large runtime component when I perfed/flamegraphed this.

# Timing

### Baseline
gray: 70ms
rgb: 116ms

### With Clamping Rework
gray: 70ms -> 51ms
rgb: 116 -> 93ms

### New LUT
gray: 38ms
rgb: 79ms

### New LUT + Clamping Rework
gray: 38ms -> 17ms
rgb: 79ms -> 59ms

## 0.25 LUT-based gray only
gray: 70ms

### DIFF
--- a/src/filter/bilateral.rs
+++ b/src/filter/bilateral.rs
@@ -1,6 +1,6 @@
 //! Bilateral Filter and associated items.
 
-use image::{GenericImage, Pixel};
+use image::{GenericImage, Luma, Pixel, Rgb, Rgba};
 use num::cast::AsPrimitive;
 
 use crate::definitions::Image;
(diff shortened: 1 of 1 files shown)


src/filter/bilateral.rs
```diff
@@ -138,54 +292,69 @@ where
...
         }
+    }
 
-        let mut out_pixel = center_pixel;
-        let num_channels = P::CHANNEL_COUNT as usize;
-        let out_channels = out_pixel.channels_mut();
-        for i in 0..num_channels {
-            out_channels[i] = (channel_sums[i] / weight_sum).as_();
+    // Bottom edge
+    for y in (height - radius)..height {
```
cospectrum, line 348: I think this overlaps with the previous loop of `y in 0..radius`, for example:
```
radius = 7
height = 10

0..radius <=> 0..7
(height - radius)..height <=> (10 - 7)..10 <=> 3..10
```

src/filter/bilateral.rs
```diff
@@ -56,6 +56,158 @@ where
...
+    f32: From<P::Subpixel> + AsPrimitive<P::Subpixel>,
+{
+    /// Filter a single output pixel.
+    ///
+    /// When `CLAMP` is `true` the window coordinates are clamped into the image bounds
+    /// (used for border pixels); when `false` the window is assumed fully in-bounds
+    /// (used for the interior). Because `CLAMP` is a const generic, each variant
+    /// monomorphizes to straight-line code with the unused branch eliminated.
+    #[inline(always)]
+    fn filter<const CLAMP: bool>(&self, x: u32, y: u32) -> P {
```
cospectrum, line 157: should be unsafe

src/filter/bilateral.rs
```diff
@@ -56,6 +56,158 @@ where
...
+/// A gaussian function of the euclidean distance between two pixel's colors, implemented using
+/// a look up table. This gives substantial speed up for pixel's with a u8 subpixel.
+///
+/// This implements [`ColorDistance`].
+pub struct LutGaussianEuclideanColorDistance {
+    lut: Vec<f32>,
+}
+
+impl LutGaussianEuclideanColorDistance {
+    /// Creates a new [`GaussianEuclideanColorDistance`] using a given sigma value, which
```
cospectrum, line 68: [`LutGaussianEuclideanColorDistance`]

src/filter/bilateral.rs
```diff
@@ -56,6 +56,158 @@ where
...
+    ///
+    /// Internally, this is stored as sigma squared for performance.
+    ///
+    /// # Panics
+    ///
+    /// 1. If `sigma <= 0`
+    pub fn new(sigma: f32) -> Self {
+        assert!(
+            sigma > 0.0,
+            "GaussianEuclideanColorDistance sigma must be positive"
```
cospectrum, line 79: LutGaussianEuclideanColorDistance

src/filter/bilateral.rs
```diff
@@ -56,6 +56,158 @@ where
     }
 }
 
+/// A gaussian function of the euclidean distance between two pixel's colors, implemented using
+/// a look up table. This gives substantial speed up for pixel's with a u8 subpixel.
+///
+/// This implements [`ColorDistance`].
+pub struct LutGaussianEuclideanColorDistance {
```
cospectrum, line 63: needs testing

src/filter/bilateral.rs
cospectrum: and needs testing

src/filter/bilateral.rs
cospectrum: This will do
```rust
debug_assert!(radius <= height)
let interior_end = (height - radius).max(radius);

for y in 0..radius {}

for y in radius..interior_end {}

for y in interior_end..height {}
```

src/filter/bilateral.rs
```diff
@@ -138,54 +292,69 @@ where
...
+                out_image.unsafe_put_pixel(x, y, val);
+            }
+        }
 
-                weight_sum += weight;
-                for (i, c) in window_pixel.channels().iter().enumerate() {
-                    channel_sums[i] += weight * f32::from(*c);
-                }
+        // Right edge
+        for x in (width - radius)..width {
```
cospectrum, line 339: overlaps similar to comment below

src/filter/bilateral.rs
```diff
@@ -56,6 +56,158 @@ where
...
+                    channel_sums[i] += weight * f32::from(*c);
+                }
+            }
+        }
+
+        let mut out_pixel = center_pixel;
+        let num_channels = P::CHANNEL_COUNT as usize;
+        let out_channels = out_pixel.channels_mut();
+        for i in 0..num_channels {
+            out_channels[i] = (channel_sums[i] / weight_sum).as_();
```
cospectrum, line 205: Probably faster with zip because `out_channels` is a slice, plus will not go out of bounds

src/filter/bilateral.rs
```diff
@@ -56,6 +56,158 @@ where
...
+    ///
+    /// # Panics
+    ///
+    /// 1. If `sigma <= 0`
+    pub fn new(sigma: f32) -> Self {
+        assert!(
+            sigma > 0.0,
+            "GaussianEuclideanColorDistance sigma must be positive"
+        );
+        let mut lut = Vec::with_capacity(512);
```
cospectrum, line 81: maybe lets allocate `[f32; 512]` on the stack?

cospectrum (changes requested):

src/filter/bilateral.rs
```diff
@@ -124,7 +276,9 @@ where
     assert_ne!(image.height(), 0);
     assert!(spatial_sigma > 0.0, "spatial_sigma must be positive");
 
-    let radius = i16::from(radius);
+    let (width, height) = image.dimensions();
+    let radius = (radius as u32).min(width).min(height);
```
cospectrum, line 280: I think we should document that `radius` is clamped to be in bounds

cospectrum (changes requested):

